PR #1, opened 2021-08-19
Feature/pmb59 testing

pmb59:
WIP PR - addresses some of the issues and errors found so far

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -127,7 +127,7 @@ def get_outputs():
     print(outputs)
     print('Getting list of outputs.. done')
     
-    if config['delete_previous_output']==True:
+    if 'delete_previous_output' in config and config['delete_previous_output']==True:
         for x in outputs:
             print('Trying to delete existing output: '+ x)
             try:
--- a/Snakefile
+++ b/Snakefile
@@ -126,12 +126,14 @@ def get_outputs():
         outputs.extend(expand(f"{config['accession']}"+"-{metric}-coexpressions.tsv.gz", metric=get_metrics() )) 
     print(outputs)
     print('Getting list of outputs.. done')
-    for x in outputs:
-        print('Trying to delete existing output: '+ x)
-        try:
-            os.remove(x)
-        except:
-            print("Output file ", x, " not found in ", os.getcwd())
+    
+    if config['delete_previous_output']==True:
+        for x in outputs:
+            print('Trying to delete existing output: '+ x)
+            try:
+                os.remove(x)
+            except:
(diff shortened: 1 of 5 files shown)


Snakefile
```diff
@@ -121,8 +121,18 @@ def get_outputs():
     if 'atlas-experiment-summary' in config['tool'] or 'all' in config['tool']:
         outputs.append(f"{config['accession']}-atlasExperimentSummary.Rdata")
     if 'baseline-heatmap' in config['tool'] or 'all-baseline' in config['tool']:
-        outputs.extend(expand(f"{config['accession']}"+"-{metric}.tsv", metric=get_metrics() ))
+        outputs.extend(expand(f"{config['accession']}"+"-heatmap-{metric}.pdf", metric=get_metrics() ))
+    if 'baseline-coexpression' in config['tool'] or 'all-baseline' in config['tool']:   
+        outputs.extend(expand(f"{config['accession']}"+"-{metric}-coexpressions.tsv.gz", metric=get_metrics() )) 
     print(outputs)
+    print('Getting list of outputs.. done')
+    for x in outputs:
```
pcm32: This deletion should be optional, so that on runs that happen after fixing errors in data, files that have been already computed correctly don't need to be redone. I would suggest then that we only go through the deletion part only when config holds some special flag, like `--config delete_previous_results=true` or alike.

pcm32 (commented):
Some minor comments to improve this.

Snakefile
```diff
@@ -121,8 +121,20 @@ def get_outputs():
...
         outputs.append(f"{config['accession']}-atlasExperimentSummary.Rdata")
     if 'baseline-heatmap' in config['tool'] or 'all-baseline' in config['tool']:
-        outputs.extend(expand(f"{config['accession']}"+"-{metric}.tsv", metric=get_metrics() ))
+        outputs.extend(expand(f"{config['accession']}"+"-heatmap-{metric}.pdf", metric=get_metrics() ))
+    if 'baseline-coexpression' in config['tool'] or 'all-baseline' in config['tool']:   
+        outputs.extend(expand(f"{config['accession']}"+"-{metric}-coexpressions.tsv.gz", metric=get_metrics() )) 
     print(outputs)
+    print('Getting list of outputs.. done')
+    
+    if config['delete_previous_output']==True:
```
```suggestion
    if 'delete_previous_output' in config and config['delete_previous_output']==True:
```
otherwise it might complain if the flag is not present

pcm32 (approved):
LGTM

PR #2, opened 2021-08-20
detele YAML if gff file is not available

pmb59:
If the gff file is not available this should now trigger a failure of the workflow.

### DIFF
--- a/Snakefile-sorting-hat
+++ b/Snakefile-sorting-hat
@@ -128,6 +128,15 @@ rule get_experiment_metadata:
         echo "gff: $gff_file" >> {output.metadata}
         echo "idf: $idf" >> {output.metadata}
         echo "sdrf: $sdrf" >> {output.metadata}
+
+        gff_found=$( find {input.gtf_dir}/$organism -iname "$organism.*.gff3" | head -n1 | wc -l)
+        if [[ $gff_found -eq 1 ]]
+        then
+            echo "gff found"
+        else
+            echo "gff not found. removing yaml..."
+            rm {output.metadata}
+        fi
         """
 
 rule produce_recalculations_call:
--- a/Snakefile-sorting-hat
+++ b/Snakefile-sorting-hat
@@ -108,31 +108,31 @@ rule get_experiment_metadata:
         export ATLAS_PROD={params.atlas_prod}
         export ATLAS_EXPS={params.atlas_exps}
         source {workflow.basedir}/bin/metadata_retrieval_functions.sh
-        perl {workflow.basedir}/bin/get_exptype_contrasts_assays.pl {wildcards.accession} {input.config_file} {output.metadata}
-        if [ -e "{wildcards.accession}/{wildcards.accession}.condensed-sdrf.tsv" ]; then
-           echo "Trying with condensed..."
-           organism=$(get_organism_from_condensed_sdrf "{wildcards.accession}/{wildcards.accession}.condensed-sdrf.tsv")
-        elif [ -e "{wildcards.accession}/{wildcards.accession}-configuration.xml" ]; then
-           echo "Trying with perl (produces organism with capitals and spaces)"
-           organism=$(perl {workflow.basedir}/bin/get_experiment_info.pl --experiment {wildcards.accession} --xmlfile {wildcards.accession}/{wildcards.accession}-configuration.xml --organism | sed 's/ /_/g' | tr '[:upper:]' '[:lower:]')
-        else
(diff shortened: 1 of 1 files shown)


Snakefile-sorting-hat
```diff
@@ -128,6 +128,15 @@ rule get_experiment_metadata:
         echo "gff: $gff_file" >> {output.metadata}
         echo "idf: $idf" >> {output.metadata}
         echo "sdrf: $sdrf" >> {output.metadata}
+
+        gff_found=$( find {input.gtf_dir}/$organism -iname "$organism.*.gff3" | head -n1 | wc -l)
```
pinin4fjords: The checks should be places before anything is output to the yaml, so before the previous few lines. You'll also want to relocate line 111 to after the check I think.

Snakefile-sorting-hat
```diff
@@ -128,6 +128,15 @@ rule get_experiment_metadata:
...
+        if [[ $gff_found -eq 1 ]]
+        then
+            echo "gff found"
+        else
+            echo "gff not found. removing yaml..."
+            rm {output.metadata}
+        fi
```
pinin4fjords: Could you match bash conventions used elsewhere to help with readability? Also, you don't need to re-do the search, and if you do this check before any content is added to the yaml, then you won't need to remove it.

I suggest this 
```suggestion
        if [ $(echo "$gff_found" | wc -l) -ne 1 ]; then
            echo "gff not found." 1>&2
            exit 1
        fi
```

pmb59:
fixed accordingly, TBR

Snakefile-sorting-hat
```diff
@@ -108,26 +108,32 @@ rule get_experiment_metadata:
...
+                organism=$(perl {workflow.basedir}/bin/get_experiment_info.pl --experiment {wildcards.accession} --xmlfile {wildcards.accession}/{wildcards.accession}-configuration.xml --organism | sed 's/ /_/g' | tr '[:upper:]' '[:lower:]')
+            else
+                echo "Nothing worked..."
+                >&2 echo "Can't retrieve organism: neither {wildcards.accession}/{wildcards.accession}.condensed-sdrf.tsv nor {wildcards.accession}/{wildcards.accession}-configuration.xml not found "
+                exit 1
+            fi
+            magetabfiles=$( perl {workflow.basedir}/bin/get_experiment_info.pl --experiment {wildcards.accession} --xmlfile {wildcards.accession}/{wildcards.accession}-configuration.xml --magetabfiles)
+            idf=$( echo $magetabfiles | awk -F',' '{{ print $1 }}' )
+            sdrf=$( echo $magetabfiles | awk -F',' '{{ print $2 }}' )
+            echo "organism: $organism" >> {output.metadata}
```
pcm32, line 135: Maybe we should check in the same way and exit with 1 if organism, idf or sdrf files are not well defined, as that would mean a failure later on in the analysis parts.

Snakefile-sorting-hat
```diff
@@ -108,26 +108,32 @@ rule get_experiment_metadata:
...
         magetabfiles=$( perl {workflow.basedir}/bin/get_experiment_info.pl --experiment {wildcards.accession} --xmlfile {wildcards.accession}/{wildcards.accession}-configuration.xml --magetabfiles)
         idf=$( echo $magetabfiles | awk -F',' '{{ print $1 }}' )
         sdrf=$( echo $magetabfiles | awk -F',' '{{ print $2 }}' )
-        gff_file=$( find {input.gtf_dir}/$organism -iname "$organism.*.gff3" | head -n1)
-        echo "organism: $organism" >> {output.metadata}
-        echo "gff: $gff_file" >> {output.metadata}
-        echo "idf: $idf" >> {output.metadata}
-        echo "sdrf: $sdrf" >> {output.metadata}
+        if [ -z "${gff_file}" ] || [ -z "${idf}" ] || [ -z "${sdrf}" ]; then
+            echo "gff, idf or sdrf not found. Yaml not created"  1>&2
```
```suggestion
            echo "GFF: $gff_file"
            echo "IDF: $idf"
            echo "SDRF: $sdrf"
            echo "GFF, IDF or SDRF not found. YAML not created"  1>&2
```
Not sure if we need the 1>&2 here... since other echos are just going to stdout and we are redirecting everything at the top...

pcm32: I added those so that we know what is missing....

pcm32 (approved):
LGTM, please decide what to do with that echo redirection and whether we want to add the field printing suggested to know what is missing if something is.

PR #3, opened 2021-08-24
Deals with changed path after pushd

pcm32:
A change I made in the cluster that I forgot to put on the version control copy. This deals with the YAML file not being found due to the change of relative position after the pushd.


pmb59 (approved):
LGTM

PR #4, opened 2021-08-24
Minimal python code that check if <accession>-fpkms.tsv and <accessio…

pmb59:
…n>-tpms.tsv exist before executing the recalculations.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -60,10 +60,19 @@ def get_ext_db():
         return ["go", "reactome", "interpro"]
 
 def get_metrics():
+    import glob
     if 'metric' in config:
         return config['metric'].split(":")
     else:
-        return ['tpms', "fpkms"]
+        metric_types = ['tpms', 'fpkms'] 
+        metric_grabbed = []
+        for files in metric_types:
+            metric_len = len( glob.glob(f"{config['accession']}-"+files+".tsv") )
+            #print(metric_len)
+            if metric_len==1:
+                metric_grabbed.append(files)
+        print(metric_grabbed)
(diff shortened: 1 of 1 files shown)


Snakefile
```diff
@@ -60,10 +60,19 @@ def get_ext_db():
...
-        return ['tpms', "fpkms"]
+        metric_types = ['tpms', 'fpkms'] 
+        metric_grabbed = []
+        for files in metric_types:
+            metric_len = len( glob.glob(f"{config['accession']}-"+files+".tsv") )
+            #print(metric_len)
+            if metric_len==1:
+                metric_grabbed.append(files)
+        print(metric_grabbed)
+        return metric_grabbed     # ideally: ['tpms', "fpkms"]
```
```suggestion 
        metric_grabbed = []
        for metric in ['tpms', 'fpkms']:
            # glob returns a list of matching paths, so if the metric is available a non-empty list is returned.
            if glob.glob(f"{config['accession']}-{metric}.tsv"):
                metric_grabbed.append(metric)
        return metric_grabbed     # ideally: ['tpms', "fpkms"]
```
A bit of simplification on the code:
- files is a bit misleading as a variable, also should be singular if an iterating element of a list
- iterate directly on the list, since it won't be used again.
- Since glob returns a possibly empty list, you can do if glob_result and if the list is empty it evaluates to false on the if.

This should work, but please test if possible.

pcm32 (approved):
Idea looks good, I have simplified a bit the code, please test before merging.

pmb59:
The simplified version runs well.

PR #5, opened 2021-08-25
Feature/skip steps

pmb59:
Changes to address rule skipping for specific experiments in bulk recalculations - as specified in a YAML file.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -120,7 +120,7 @@ def get_outputs():
     # Read this now so that it is available for all other needs
     read_metadata_summary()
     global skip_accession
-    skip_accesion = read_skip_steps_file()
+    skip_accession = read_skip_steps_file()
     required_config=['tool']
     check_config_required(fields=required_config)
     if 'percentile-ranks' in config['tool'] or config['tool']=="all-diff" and skip(config['accession'],'percentile_ranks'):
--- a/Snakefile
+++ b/Snakefile
@@ -11,6 +11,19 @@ def read_metadata_summary():
         with open(config['metadata_summary'], 'r') as fh:
             metadata_summary = yaml.safe_load(fh)
 
+def read_skip_steps_file():
+    if 'skip_steps_file' in config:
+        global skip_steps
+        with open(config['skip_steps_file'], 'r') as stream:
+            try:
+                skip_steps=yaml.safe_load(stream)
+            except yaml.YAMLError as exc:
+                print(exc)
+        return skip_steps
+    else:
+        return [" "]
+
+
 def get_from_config_or_metadata_summary(label):
     if label in config:
         return config[label]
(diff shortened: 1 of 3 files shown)


Snakefile
```diff
@@ -11,6 +11,19 @@ def read_metadata_summary():
...
+    if 'skip_steps_file' in config:
+        global skip_steps
+        with open(config['skip_steps_file'], 'r') as stream:
+            try:
+                skip_steps=yaml.safe_load(stream)
+            except yaml.YAMLError as exc:
+                print(exc)
+        return skip_steps
+    else:
+        return [" "]
```
pcm32: just load data into dictionary, no need for returning it, as it is done with the metadata file

Snakefile-sorting-hat
```diff
@@ -171,6 +175,7 @@ rule produce_recalculations_call:
             metadata_summary=../{input.metadata} \
             bioentities_properties={params.bioentities_properties} \
             delete_previous_output={params.delete_previous_output} \
+            skip_steps_file={params.skip_steps_file} \
```
pcm32, line 185: this should happen only if provided, so this line shouldn't alway appear.

Snakefile
```diff
@@ -102,33 +115,34 @@ def get_outputs():
 
     # Read this now so that it is available for all other needs
     read_metadata_summary()
+    skip_accesion = read_skip_steps_file()
     required_config=['tool']
     check_config_required(fields=required_config)
-    if 'percentile-ranks' in config['tool'] or config['tool']=="all-diff":
+    if 'percentile-ranks' in config['tool'] or config['tool']=="all-diff" and config['accession'] not in skip_accesion['skips']['percentile-ranks']:
```
pcm32: on this sort of checks you need to protect from the case where skip_accession is empty (you will get an error due to missing key). I would write something like:

```
def skip(accession, tool):
    # if all the keys exist in skip_accession, check for the accession inside the tool part
    # if it has to skip:
    return true
```

and then call that everywhere (will make statements shorter as well).

Snakefile-sorting-hat
```diff
@@ -68,6 +68,9 @@ def get_bioentities_property_path():
 def get_delete_previous_output():
     return config['delete_previous_output']
 
+def get_skip_steps_file():
+    return config['skip_steps_file']
```
pcm32: again, consider the case when this key is not defined. Probably same applies to 'delete_previous_output' above.

pcm32 (changes requested):
Thanks for this. Some minor suggestions to improve this.

Snakefile
pmb59: Returning empty list if yaml is absent was suitable for `not in skip_accesion['skips']`. In any case, I'll remove it

pmb59:
NOTE: I've fixed in the last 2 commits a couple of issues related to the GFF.

pcm32 (approved):
Great work! Thanks for addressing my concerns.

PR #7, opened 2021-08-26
Aims to fix conda envs issue.

pmb59:


pcm32:
I made this comment elsewhere, but for you convenience:

you could get rid of all the python mumbo-jumbo and just have something like:

```
rule env_1:
   env: envs/env1.yaml
   shell:
     """
     echo "I'm easy 1"
     """

rule env_2:
   env: envs/env2.yaml...
```

and then run that as dry run with the conda creation flag.

pcm32:
...mmm... or even better, use the python API from Snakemake and iterate over all existing yamls in envs and send those to build (if they don't exist ideally)... then there is no tracking needed of new YAMLs added. Shortcoming of this could be re-generating (and failing) of existing working conda envs (in time, this tends to happen, that is why in another topics we tend to use singularity).

pcm32:
I think that this PR is obsolete compared to what you have currently, right @pmb59 ?

Snakefile
```diff
@@ -201,8 +201,8 @@ rule percentile_ranks:
             # remove only microarray derived multiple percentile ranks
             # (per array design <accession>_<arraydesign>-percentile-ranks.tsv)
             rm -f {wildcards.accession}_*-percentile-ranks.tsv
-        else
-            mv ${{percentile_ranks[0]}} {output.percentile_ranks_merged}
+            #else
+            #mv ${{percentile_ranks[0]}} {output.percentile_ranks_merged}
```
pcm32: Lets just remove this. Although strictly speaking, I would suggest a separate PR for this, then it is easier to revert if needed (and safer to remove the comments).

run_sorting_hat_test_data.sh
```diff
@@ -10,22 +10,22 @@ SN_CONDA_PREFIX=${SN_CONDA_PREFIX:-$( pwd )/conda_installs}
 CONDA_PREFIX_LINE="--conda-prefix $SN_CONDA_PREFIX"
 export LOG_PATH=${LOG_PATH:-$( pwd )/sorting.log}
 USUAL_SM_ERR_OUT=${USUAL_SM_ERR_OUT:-$( pwd )/snakemake.log}
-pushd $1
 
+#create conda envs only
+for yaml_file in $(ls $( pwd )/envs); do
+  echo $yaml_file
+  snakemake --dry-run --use-conda --conda-create-envs-only $CONDA_PREFIX_LINE  \
```
pcm32: I think that --dry-run is redundant when using --conda-create-envs-only, on the successful run (see slack 02/09/2021 ~2:44) I didn't use --dry-run.

Snakefile-sorting-hat
```diff
@@ -165,12 +165,15 @@ rule produce_recalculations_call:
...
-        snakemake {config['sm_options']} --config \
+
+
+        snakemake --latency-wait 100 {config['sm_options']} --config \
             accession={wildcards.accession} \
             tool={get_tool(wildcards.accession)} \
             metadata_summary=../{input.metadata} \
             bioentities_properties={params.bioentities_properties} \
             delete_previous_output={params.delete_previous_output} \
+            skip_steps_file={params.skip_steps_file} \
```
pcm32, line 176: Could you remind me what was this about? was it to skip steps for datasets where we don't have raw sequences?

Snakefile
pcm32: But given the small change, lets just leave it here.

pcm32 (approved):
LGTM, left a couple of minor comments, feel free to merge after addressing them.

Snakefile
pmb59: Done

run_sorting_hat_test_data.sh
pmb59: done

Snakefile-sorting-hat
pmb59: This is regarding a non-public YAML file with a reduced list of experiments that specifically can skip certain steps. This is necessary in absence of certain files.

PR #8, opened 2021-08-26
Filters out ACC based on accessions that are available in atlas_exps

pmb59:


Snakefile-sorting-hat
```diff
@@ -79,12 +79,22 @@ def get_skip_steps_file():
...
         ACC=config['accessions'].split(":")
     else:
         ACC=glob.glob('E-*')
-    return expand("{accession}/{accession}.recalculations.done", accession=ACC)
+
+    if 'atlas_exps' in config:  
+        # filter out ACC based on accessions that are available in atlas_exps
+        pass_ACC = []
+        for x in ACC:
+            if x in next(os.walk( config['atlas_exps'] ))[1]:
```
pcm32: Making an os system call to list the atlas_exps directory every time will be very expensive on the shared file system. Please retrieve first the complete list of files in the that directory and then compare each accession against that. I would suggest as well that we don't use the atlas_exps Atlas jargon here and just call that `filtering_directory` or `acceptable_accessions_dir` or something that conveys that the directory is being used for here.

pcm32: We agreed to ignore the atlas_exps, atlas_prod jargon usage for now.

Snakefile-sorting-hat
```diff
@@ -79,12 +79,23 @@ def get_skip_steps_file():
...
         ACC=glob.glob('E-*')
-    return expand("{accession}/{accession}.recalculations.done", accession=ACC)
+
+    if 'atlas_exps' in config:  
+        # filter out ACC based on accessions that are available in atlas_exps
+        pass_ACC = []
+        valid_acc_dir = next(os.walk( config['atlas_exps'] ))[1]
+        for x in ACC:
+            if x in valid_acc_dir:
+                pass_ACC.append(x)
```
```suggestion
        valid_acc_dir = next(os.walk( config['atlas_exps'] ))[1]
        # intersection of the two lists
        pass_ACC = list(set(ACC) & set(valid_acc_dir))
```

pcm32 (approved):
Looks good, a minor suggestion to simplify code, please merge after committing that if it looks good.

Snakefile-sorting-hat
pcm32: I think that `set(ACC).intersection(valid_acc_dir)` would also work and be more explicit.

pcm32: but might require a list cast though...

pmb59: LGTM. Less code and it should run quicker than the for loop.

PR #9, opened 2021-08-27
WIP Saves to file experiments that actually pass the filter. Useful for d…

pmb59:
…ebugging

### DIFF
--- a/Snakefile-sorting-hat
+++ b/Snakefile-sorting-hat
@@ -91,6 +91,10 @@ def get_outputs():
         valid_acc_dir = next(os.walk( config['atlas_exps'] ))[1]
         # intersection of the two lists
         pass_ACC = list(set(ACC) & set(valid_acc_dir))
+        #save passing ACCs to file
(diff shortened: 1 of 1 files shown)


Snakefile-sorting-hat
```diff
@@ -91,6 +91,10 @@ def get_outputs():
         valid_acc_dir = next(os.walk( config['atlas_exps'] ))[1]
         # intersection of the two lists
         pass_ACC = list(set(ACC) & set(valid_acc_dir))
+        #save passing ACCs to file
+        with open('exps_tagged_for_processing.txt', 'w') as f:
```
pcm32, line 95: looks good, but maybe you want to set the file path here from config, otherwise it will end up in the analysis directory from where it is getting experiments.

pcm32 (approved):
Feel free to merge after addressing the path for the file.

pmb59:
This is an old PR, and was used only for debugging.

PR #10, opened 2021-08-31
Check and fix -if necessary- species names for bulk recalculations

pmb59:
WIP PR.


pcm32:
There are more cases of species redirections that need to happen, so I would try to favour something that can be maintained by the curators externally (possibly a species name mapping file in one of the metadata repos). Then we could add some logic on these pipelines, so that if provided (as a config element possibly), the pipeline checks the contents of such a mappings file and uses any mappings added there. In the production environment, the pipeline would be configured to read this file from a specific checkout of the repo that would contain such mapping file. Then we could also redirect older Atlas setup to use that maybe and completely remove then things like [this](https://github.com/ebi-gene-expression-group/atlas-prod/blob/d9fba864273b7a5b79cdbbf8822e5e566b07fbaa/analysis/gsa/scripts/gsa_prepare_data.sh#L23) and others.

pcm32:
What do you think @anjaf , where would you keep such a file with species name mappings within the curators repos? Thanks.

anjaf:
Would this fit into the [atlas-config](https://gitlab.ebi.ac.uk/ebi-gene-expression/atlas-config) repo? 

I also found a few similar files listed here: https://github.com/ebi-gene-expression-group/perl-atlas-modules/blob/develop/supporting_files/AtlasSiteConfig.yml.default 
(not sure though where they would be, e.g. scattered around the file system or part of other repos already?) 
Perhaps it would make sense to also move all of those files to the config repo too.

pcm32:
@pmb59 could you propose a file structure for the species mapping on the repo https://gitlab.ebi.ac.uk/ebi-gene-expression/atlas-config and then add something here to read such a file (like a config option). Then on the cluster we make sure that the config points to the file in an adequate checkout of that repo.

pcm32:
Or were you thinking to place this in a specific file @anjaf ? Maybe there is already somewhere there a YAML where we could add a species mapping dictionary?

anjaf:
My knowledge of existing Atlas species mappings is limited. I would create a new file for now.

pmb59:
@pcm32 @anjaf Is the consensus to create a new YAML at https://gitlab.ebi.ac.uk/ebi-gene-expression/atlas-config with those mappings (incorrect sp. name -> correct sp. name)? Where can I find a list of all of those cases? Or is it mostly scattered across different repositories?

Snakefile-sorting-hat
```diff
@@ -142,6 +150,16 @@ rule get_experiment_metadata:
             >&2 echo "Can't retrieve organism: neither {wildcards.accession}/{wildcards.accession}.condensed-sdrf.tsv nor {wildcards.accession}/{wildcards.accession}-configuration.xml not found "
             exit 1
         fi
+        # fix species name
+        #organism_check=$( python3  {workflow.basedir}/bin/bulk_recalculations_check_species_names.py $(echo $organism) {params.check_species_file} )
```
```suggestion
```

pcm32: I would remove this comment.

Snakefile-sorting-hat
```diff
@@ -142,6 +150,16 @@ rule get_experiment_metadata:
...
+        while read line; do
+        echo "$line"
+        DIR=$line
+        done < <(perl {workflow.basedir}/bin/bulk_recalculations_check_species_names.pl $(echo $organism) {params.check_species_file}  )
+	    organism_check=$(echo $DIR)
+
+        echo $organism "check-> "$organism_check >> organisms.txt
```
pcm32: Do we need this organisms.txt file there?

bin/bulk_recalculations_check_species_names.pl
```diff
@@ -0,0 +1,53 @@
...
+
+#Search
+my $found = 0;
+my $output_sp = ' ';
+
+for my $k1 (keys %{$data}) {
+    for my $k2 (keys %{$data->{$k1}}) {
+        #print "Key:    $k2 => @{$data->{$k1}{$k2}}\n";
+        #print "$k2\n";
+        #print "@{$data->{$k1}{$k2}} \n";
```
pcm32:
```suggestion
```

bin/bulk_recalculations_check_species_names.pl
```diff
@@ -0,0 +1,53 @@
...
+        # transform arrays into strings
+        my $string1 = join('-' ,$k2);
+        my $string2 = join( '-', @{$data->{$k1}{$k2}} );
+        print "Key:  '$string1' -> '$string2' \n";
+        #print "Input $input_sp \n";
```
pcm32:
```suggestion
```

bin/bulk_recalculations_check_species_names.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python3
```
pcm32: I would get rid of this script if we are not using it. Or at least clarify that we are not using it.

pcm32 (approved):
Added some minor cleanup suggestions, but LGTM in general.

Snakefile-sorting-hat
pmb59: Not really, only useful for testing. I'll comment it for now

bin/bulk_recalculations_check_species_names.py
pmb59: Done

PR #12, opened 2021-09-15
Fixes on rule link_baseline_coexpression

pmb59:


pinin4fjords:
@pmb59 I've applied a putative fix for the symlinking once I understood the logic better. It will need testing but you should see what I mean. Basically we can just use what we already know from get_metrics() to make that rule work. I also made it so get_metrics() is only run once so we don't keep globbing the file system (naughty @pcm32).

Snakefile
```diff
@@ -82,6 +82,7 @@ def get_metrics():
                 metric_grabbed.append(metric)
```
pinin4fjords, line 85: We should add a check here that at least one metric is available

Snakefile
```diff
@@ -146,9 +150,10 @@ def get_outputs():
     if 'atlas-experiment-summary' in config['tool'] or 'all' in config['tool'] and skip(config['accession'],'atlas_experiment_summary'):
         outputs.append(f"{config['accession']}-atlasExperimentSummary.Rdata")
     if 'baseline-heatmap' in config['tool'] or 'all-baseline' in config['tool'] and skip(config['accession'],'baseline-heatmap'):
-        outputs.extend(expand(f"{config['accession']}"+"-heatmap-{metric}.pdf", metric=get_metrics() ))
+        outputs.extend(expand(f"{config['accession']}"+"-heatmap-{metric}.pdf", metric=metrics ))
     if 'baseline-coexpression' in config['tool'] or 'all-baseline' in config['tool'] and skip(config['accession'],'baseline-coexpression'):   
-        outputs.extend(expand(f"{config['accession']}"+"-{metric}-coexpressions.tsv.gz", metric=get_metrics() )) 
+        outputs.extend(expand(f"{config['accession']}"+"-{metric}-coexpressions.tsv.gz", metric=metrics ))
```
pinin4fjords: This needs to be modified to check the number of columns in the expression files (TPM and FPKM should have the same number of columns), and only request the coexpression files if the column number is greater than 3. 

This is a lot simpler to do than trying to account for dynamic failures in the baseline_coexpression rule.

pinin4fjords: FYI @pcm32

pmb59: Do you mean simpler than using a checkpoint in `baseline_coexpression`?

pinin4fjords: Checkpoint is overly complex for situations where we know in advance the output files. It's very easy here to read the first line of the TSVs  for FPKM and TPM and decide based on that whether or not we can execute the coexpression steps. It would also be wasteful of resources to submit cluster jobs for this when we know in advance they will fail.

Snakefile
pmb59: See  d84a8a2

Snakefile
```diff
@@ -329,22 +348,14 @@ rule link_baseline_coexpression:
...
         exec &> "{log}"
-        if [ -s {wildcards.accession}-tpm-coexpressions.tsv.gz ]; then
-            ln -s {wildcards.accession}-tpm-coexpressions.tsv.gz {wildcards.accession}-coexpressions.tsv.gz
-        elif [ -s {wildcards.accession}-fpkm-coexpressions.tsv.gz ]; then
-            ln -s {wildcards.accession}-fpkm-coexpressions.tsv.gz {wildcards.accession}-coexpressions.tsv.gz
-        else
-            echo "Error: neither TPM nor FPKM coexpressions.tsv.gz file found"
-            exit 1
-        fi
+	    ln -s {input} {output}
```
```suggestion
	    ls -l {input)
	    ln -s {input} {output}
	    ls -l {input}
	    ls -l {output}
```

Snakefile
```diff
@@ -146,9 +162,17 @@ def get_outputs():
     if 'atlas-experiment-summary' in config['tool'] or 'all' in config['tool'] and skip(config['accession'],'atlas_experiment_summary'):
         outputs.append(f"{config['accession']}-atlasExperimentSummary.Rdata")
     if 'baseline-heatmap' in config['tool'] or 'all-baseline' in config['tool'] and skip(config['accession'],'baseline-heatmap'):
-        outputs.extend(expand(f"{config['accession']}"+"-heatmap-{metric}.pdf", metric=get_metrics() ))
+        outputs.extend(expand(f"{config['accession']}"+"-heatmap-{metric}.pdf", metric=metrics ))
     if 'baseline-coexpression' in config['tool'] or 'all-baseline' in config['tool'] and skip(config['accession'],'baseline-coexpression'):   
-        outputs.extend(expand(f"{config['accession']}"+"-{metric}-coexpressions.tsv.gz", metric=get_metrics() )) 
+        metric_link_coexp=False
+        for m in metrics:
+            if os.path.getsize(f"{config['accession']}"+"-"+m+".tsv") > 0 and get_number_columns(f"{config['accession']}"+"-"+m+".tsv")>4:
```
```suggestion
            expression_file=f"{config['accession']}-{m}.tsv"
            print(f"Checking file size for {expression_file} and number of columns")
            if os.path.getsize(expression_file) > 0 and get_number_columns(expression_file)>4:
```

Snakefile
```diff
@@ -146,9 +162,17 @@ def get_outputs():
...
+                metric_link_coexp=True
+                outputs.extend(expand(f"{config['accession']}"+"-"+m+"-coexpressions.tsv.gz"))
```
```suggestion
                outputs.extend(expand(f"{config['accession']}-{m}-coexpressions.tsv.gz"))
```

Snakefile
```diff
@@ -329,22 +362,17 @@ rule link_baseline_coexpression:
...
-        elif [ -s {wildcards.accession}-fpkm-coexpressions.tsv.gz ]; then
-            ln -s {wildcards.accession}-fpkm-coexpressions.tsv.gz {wildcards.accession}-coexpressions.tsv.gz
-        else
-            echo "Error: neither TPM nor FPKM coexpressions.tsv.gz file found"
-            exit 1
-        fi
+        ls -l {input}
+        ln -s {input} {output}
+        ls -l {input}
+        ls -l {output}
```
```suggestion
        ln -s {input} {output}
```

we can remove these checks and go back to this now.

Snakefile-sorting-hat
```diff
@@ -187,7 +187,7 @@ rule produce_recalculations_call:
         pushd {input.working_directory}
 
 
-        snakemake --latency-wait 100 {config['sm_options']} --config \
+        snakemake --forceall --latency-wait 150 {config['sm_options']} --config \
```
pcm32: `--forceall` needs to be optional, so that if a variable is set, then it runs with it. We need to be able to turn this on and off. It will only be used like this on the first run of the release, then subsequent runs of the same release shouldn't have to re-run everything.

Snakefile
```diff
@@ -146,20 +174,30 @@ def get_outputs():
...
-    
-    if 'delete_previous_output' in config and config['delete_previous_output']==True:
+    print(datetime.datetime.now())
+    if read_delete_previous_output():
         for x in outputs:
             print('Trying to delete existing output: '+ x)
             try:
                 os.remove(x)
             except:
                 print("Output file ", x, " not found in ", os.getcwd())
```
```suggestion
```

If using `--forceall` this can go now.

run_sorting_hat_test_data.sh
```diff
@@ -1,12 +1,20 @@
...
 BIOENTITIES_PROPERTIES=$( pwd )/test-data/bioentity_properties
 SORTING_HAT=$( pwd )/Snakefile-sorting-hat
 LOG_HANDLER=$( pwd )/log_handler.py
 SN_CONDA_PREFIX=${SN_CONDA_PREFIX:-$( pwd )/conda_installs}
+if [ $FORCEALL ]
+then
+  PROFILE_LINE="--profile profilename --restart-times "$RESTART_TIMES
+else
+  PROFILE_LINE="--profile profilename --forceall --restart-times "$RESTART_TIMES
+fi
```
pcm32: I wouldn't mix all these things together here... if forceall should only change a $FORCEALL_LINE or something alike.

Snakefile-sorting-hat
```diff
@@ -190,12 +183,11 @@ rule produce_recalculations_call:
         pushd {input.working_directory}
 
 
-        snakemake --forceall --latency-wait 150 {config['sm_options']} --config \
+        snakemake --latency-wait 150 {config['sm_options']} --config \
```
pcm32, line 206: How are we passing the forceall behaviour to the Snakefile analysis part? When the sorting hat receives the command to force, then the subworkflow needs to apply it as well, otherwise older results (from a previous release, won't be re-written).

run_sorting_hat_test_data.sh
```diff
@@ -1,12 +1,20 @@
 #!/usr/bin/env bash
 
 GTF=$( pwd )/test-data/gff
-DELETE_PREV_OUTPUT=False
+FORCEALL=$true
+RESTART_TIMES=1
 SKIP_STEPS=$( pwd )/step_skip.yaml
+CHECK_SPECIES=$( pwd )/Bulk_recalculations_check_species_names.yaml
```
pcm32: A better name for this file would be `atlas-species-name-mapping.yaml`, since it won't be used only on recalculations.

pcm32 (commented):
Some comments to help improve this for longer term usage and adoption within the overall processes.

Snakefile-sorting-hat
pmb59: In the current implementation, `--forceall` is passed (if `FORCEALL=$true`) here `{config['sm_options']}`  to the snakemake call in the sorting hat. See ddce3e9

run_sorting_hat_test_data.sh
pmb59: See c44db20

run_sorting_hat_test_data.sh
pmb59: See ddce3e9

run_sorting_hat_test_data.sh
```diff
@@ -1,20 +1,20 @@
 #!/usr/bin/env bash
 
+NEXPS=${NEXPS:-30}
+NJOBS=${NJOBS:-10}
 GTF=$( pwd )/test-data/gff
-FORCEALL=$true
+FORCEALL=true
```
pcm32, line 6:
```suggestion
FORCEALL=${FORCEALL:-false}
```
I would leave the least dangerous case as default.

pcm32 (approved):
LGTM

run_sorting_hat_test_data.sh
pcm32: On second thoughts, this is not part of the PR, it was probably resolved elsewhere.... I'll remove the commit suggestion. Feel free to merge as it is.

PR #13, opened 2021-09-21
Fix detection of GFF file for organisms in rule `get_experiment_metadata`

pmb59:
…tyrhynchos`, that the GFF file was not correctly captured by the logic when the summary YAML is created in bulk recalculations. This fixes the problem.

### DIFF
--- a/Snakefile-sorting-hat
+++ b/Snakefile-sorting-hat
@@ -144,6 +144,9 @@ rule get_experiment_metadata:
         fi
 
         gff_file=$( find {input.gtf_dir}/$organism -iname "$organism.*.gff3" | head -n1)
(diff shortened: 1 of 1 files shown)


pcm32 (approved):
LGTM - you might want to add some comment there explaining why we have this extra case. Very short of course.

pmb59:
For the majority of organisms at the GXA, the gtf dir contains GFF files in the form `"$organism.*.gff3"`. However, for a few organisms the GFF should be found using instead the pattern "$organism_*.gff3", like for example for `theobroma_cacao` and its GFF file `Theobroma_cacao_criollo.Criollo_cocoa_genome_V2.51.gff3`

PR #14, opened 2021-09-22
Disables checks in`check_config_required`

pmb59:
Removes 'contrast_ids' and 'organism' from `check_config_required`. Initially these would be passed through the config, but then moved away from that and left that as optional.


pmb59:
Ready to Review + S&M

pcm32 (approved):
LGTM

PR #16, opened 2021-09-28
Feature/irap gse piano

pmb59:
The modifications implemented in this feature branch resolve an `Execution halted` error found for few accessions. Now, the IRAP method `irap_GSE_piano` receives memory and threads as defined in Snakemake.

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -317,10 +317,10 @@ rule check_differential_gsea:
     """
     log: "logs/{accession}.{contrast_id}.{ext_db}-check-differential_gsea.log"
     input:
-	    gsea="{accession}.{contrast_id}.{ext_db}.gsea.tsv",
+        gsea="{accession}.{contrast_id}.{ext_db}.gsea.tsv",
         gsea_list="{accession}.{contrast_id}.{ext_db}.gsea_list.tsv"
     output:
-	    log1="logs/{accession}.{contrast_id}.{ext_db}.gsea.tsv",
+        log1="logs/{accession}.{contrast_id}.{ext_db}.gsea.tsv",
         log2="logs/{accession}.{contrast_id}.{ext_db}.gsea_list.tsv"
     shell:
         """
--- a/Snakefile
+++ b/Snakefile
@@ -268,6 +268,8 @@ rule differential_tracks:
 rule differential_gsea:
     conda: "envs/irap.yaml"
     log: "logs/{accession}.{contrast_id}.{ext_db}-differential_gsea.log"
+    resources: mem_mb=get_mem_mb
+    threads: 8
     params:
         organism=get_organism(),
         BIOENTITIES_PROPERTIES_PATH=config['bioentities_properties'],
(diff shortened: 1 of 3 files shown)


pmb59:
I have added in 4aea6e5 a new logic within the shell of `rule differential_gsea`. If the annotation file for "go", "interpro" or "reactome" db cannot be found, empty output files are generated together with a warning. This might happen if a species is not included, for instance, in the current Reactome release.

Snakefile
```diff
@@ -268,6 +268,8 @@ rule differential_tracks:
 rule differential_gsea:
     conda: "envs/irap.yaml"
     log: "logs/{accession}.{contrast_id}.{ext_db}-differential_gsea.log"
+    resources: mem_mb=get_mem_mb
+    threads: 8
```
pcm32, line 308: Does this get injected into the resources requested to the cluster? Or that would need to be done as well within the resources field above?

Snakefile
```diff
@@ -290,14 +292,21 @@ rule differential_gsea:
...
+            plotTitle="
+            Top 10 {params.ext_db_label} enriched in
+            {params.contrast_label}
+            (Fisher-exact, FDR < 0.1)"
+            {workflow.basedir}/bin/gxa_calculate_gsea.sh {wildcards.accession} $annotationFile $analyticsFile $pvalColNum $log2foldchangeColNum ./ {wildcards.contrast_id} "$plotTitle" {params.organism} {wildcards.ext_db} {threads}
+        else
+            touch {wildcards.accession}.{wildcards.contrast_id}.{wildcards.ext_db}.gsea.tsv
+            touch {wildcards.accession}.{wildcards.contrast_id}.{wildcards.ext_db}.gsea_list.tsv
+            touch {wildcards.accession}.{wildcards.contrast_id}.{wildcards.ext_db}.gsea.tsv_WARNING_not_computed
+            touch {wildcards.accession}.{wildcards.contrast_id}.{wildcards.ext_db}.gsea_list.tsv_WARNING_not_computed
```
pcm32: ...mmm... my concern here is what would happen when the web application tries to load these empty files... couldn't we try to avoid the execution of the tool (for that database) when a specific database enrichment files are not available for the organism instead? Does this happen for other combinations besides ovis_aries / reactome pair?

pcm32: maybe a finalising step could remove these if they are empty? Wouldn't checkpoints help here?

pcm32 (commented):
I have some concerns about us leaving those empty files behind...

Snakefile
pmb59: I think I've seen few cases where returned GSEA files are empty because nothing passed the significance cutoff, so it is possible the web app could handle it. But I agree that leaving empty files is not the best choice, I'll modify this

pmb59: I've added in commits 3cfca26 8490543 and 0a470b8 a new rule that checks all the output produced in`rule differential_gsea`. If the outputs are empty or have only the header (it might happen if no significant results are reported, although rarely) deletes them, and save a log files with information about the checks done on those files. These rule is helps to avoid propagating empty files to the gxa front-end.

Snakefile
```diff
@@ -304,11 +305,51 @@ rule differential_gsea:
...
+    Here we aim to delete those empty files.
+    """
+    log: "logs/{accession}.{contrast_id}.{ext_db}-check-differential_gsea.log"
+    input:
+	    gsea="{accession}.{contrast_id}.{ext_db}.gsea.tsv",
+        gsea_list="{accession}.{contrast_id}.{ext_db}.gsea_list.tsv"
+    output:
+	    log1="logs/{accession}.{contrast_id}.{ext_db}.gsea.tsv",
+        log2="logs/{accession}.{contrast_id}.{ext_db}.gsea_list.tsv"
+    shell:
```
pcm32: There are some misalignments on the indents here, please fix. Also, I would just leave all these extra logging lines on the main log file of this rule, or did you have any reason not to do it like that? If you need an output file to be generated, there is some dummy or temp mechanism and you can just make a `touch check_differential_gsea.done` in the logs folder, which will get deleted if it is a temp.

pcm32 (approved):
LGTM in general, please address the comment I added and then feel free to merge.

Snakefile
pmb59: Thanks, very useful. Please see commits below - primarily 3ba9d1f

PR #17, opened 2021-09-30
Feature/increase mem

pmb59:
Assigns (attempt x 2000) MB memory to specific rules in baseline analysis to avoid out of memory error.


pcm32 (approved):
LGTM

PR #19, opened 2021-10-01
Increases mem_mb in rule differential_tracks

pmb59:
- Increases memory in rule differential_tracks
- Renames functions that return `mem_mb` based on number of attempts


pcm32 (approved):
LGTM - you could have a single get mem method that receives the step increase as a parameter set on each rule. But let's merge this for now to avoid delaying the data release.

PR #20, opened 2021-10-13
Feature/log handler

pmb59:
Add new logic in log handler script to retrieve precise information of which rule/s is failing


pcm32 (approved):
LGTM

PR #21, opened 2021-10-22
Refactor for reprocess

pcm32:
This PR intends to refactor all rules and python code, besides get_outputs() and all rules, to a main Snakefile where we can have shared rules and logic for both reprocess and recalculate for atlas bulk. In time, the proteomics rules could also make it here.


pcm32:
LGTM - although I cannot approve because I started the PR.

pcm32:
Lets go ahead if you have tested recalculations from this branch and it is working, worst case we can revert.

PR #23, opened 2021-11-03
Copies accession contents to archive folder

pmb59:
Copies current contents to archive folder before recalculations. It happens upon successful creation of the YAML file.
It also adds a timestamp file.


pcm32 (commented):
Looks good, what if we create it without failure if already exists and we avoid deleting it? That would mean that the rsync would be faster for existing files that don’t change, like the data ones (which might be good for large datasets)

Did you take a look and see if the archive currently excludes certain files? That should be somewhere on atlas-prod codebase.

pmb59:
Closing this PR as this actually is not required for ATLAS production, only for ATLAS_EXPS.

PR #24, opened 2021-11-12
Add new experiment processing/ reprocessing rules to gxa

pmb59:
The aim of this PR is to add new rules and make of use existing ones (recalculations) to address processing of new experiments and reprocessing steps, in order to achieve a super-set of data analysis rules for GXA. This workflow now deals with previous atlas_prod scripts used before moving our systems to new EBI infrastructure.

The super-set of rules can now handle the following experiment types:
- [x] baseline
- [x] differential rnaseq
- [x] differential microarray

This PR includes: 
- logic that allows Snakemake rules collecting I/O files depending on the analysis goal (recalculations or reprocessing).
- (re)processing/recalculations only for selected specie/s
-  [updated differential expression strategy ](https://github.com/ebi-gene-expression-group/atlas-analysis/pull/1#event-6497594781)
- a number of Atlas scripts from production (which will be moved to [atlas-analysis](https://github.com/ebi-gene-expression-group/atlas-analysis) and then included as submodule here)
- communication to ISL db (only for new experiment or reprocessing)
- Major code changes/ new files are at:
```
- Snakefile
- Snakefile-reprocess
- Snakefile-sorting-hat
- bin/reprocessing_routines.sh
- bin/round_log2_fold_changes.R (reviewed)
- envs/isl-db.yaml
- run_sorting_hat_test_data.sh
```

Future merge requests will address:
- replacing generic bash routines with conda [atlas-bash-util](https://github.com/ebi-gene-expression-group/atlas-bash-util)  
- refactoring scripts using a submodule [atlas-analysis](https://github.com/ebi-gene-expression-group/atlas-analysis)
- adding proteomics and differential proteomics analysis rules
- add conditional structures that handle cases of optional/required parameters and files


Snakefile
```diff
@@ -16,9 +16,21 @@ def read_metadata_summary():
...
 read_metadata_summary()
 
-#to be implemented
-def get_isl_dir():
-    return None
+##to be implemented
+#def get_isl_dir():
+#    return None
+def get_methods_template_baseline():
+    if 'methods_base' in config:
```
pcm32, line 21:
```suggestion
    if config['methods_base']:
```
I have learnt recently that this is apparently more "pythonic".

Snakefile
```diff
@@ -94,6 +106,18 @@ def get_metrics():
         else:
             sys.exit("No metric available for baseline analyses.")
 
+def get_metrics_reprocess():
```
pcm32, line 145: Could you elaborate on why you need this distinction? in principle both of them should use the same metrics. There is no reason why you would only recalculate or reprocess only certain metrics.

Snakefile
```diff
@@ -420,16 +445,64 @@ rule atlas_experiment_summary:
 	          --output {output.rsummary}
         """
 
+
+
+# rules below are specific for reprocessing
+
+# baseline_rnaseq_experiment
+
+rule check_configuration_xml:
```
pcm32: I think that this rule can be obviated if another rules consumed the condensed SDRF explicitly.

Snakefile
```diff
@@ -420,16 +445,64 @@ rule atlas_experiment_summary:
...
+        """
+        set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
+        exec &> "{log}"
+        if ! [ -e {input} ]; then
+            exit 1
+        fi
+        touch {output}
+        """
+
+rule check_factors_xml:
```
pcm32: Idem here as in condensed

Snakefile
```diff
@@ -420,16 +445,64 @@ rule atlas_experiment_summary:
...
         raw_counts_undecorated="{accession}-raw-counts.tsv.undecorated"
     shell:
         """
-        export expIslDir={params.exp_isl_dir}
-        export expTargetDir=./
-        [ ! -z $expIslDir+x} ] || (echo "snakemake param exp_isl_dir needs to defined in rule" && exit 1)
+        set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
+        exec &> "{log}"
+        source {workflow.basedir}/bin/reprocessing_routines.sh
+        expIslDir=$(find_exp_isl_dir {wildcards.accession})
```
pcm32: I would prefer if we avoid this pattern of unearthing config mysteriously instead of passing the base ISL directory and working our way from there... if that works of course.

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+            metrictype="fpkm"
+        fi
 
-        if [ -s "$expIslDir/genes.{wildcards.metric}.htseq2.tsv" ]; then
+        if [ -s "$expIslDir/genes.$metrictype.htseq2.tsv" ]; then
             # maybe rsync could be better here?
-            cp $expIslDir/genes.{wildcards.metric}.htseq2.tsv {output.normalised_counts_undecorated}
-        elif [ -s "$expIslDir/genes.{wildcards.metric}.featurecounts.tsv" ]; then
-            cp $expIslDir/genes.{wildcards.metric}.featurecounts.tsv {output.normalised_counts_undecorated}
+            cp $expIslDir/genes.$metrictype.htseq2.tsv {output.normalised_counts_undecorated}
```
pcm32: Some of these files might be large and maybe rsync is better here?

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+        echo "ISL dir: $expIslDir"
+
+        [ ! -z $expIslDir+x ] || (echo "snakemake param exp_isl_dir needs to defined in rule" && exit 1)
+
+        if [ -s "$expIslDir/transcripts.tpm.kallisto.tsv" ] ; then
+            cp $expIslDir/transcripts.tpm.kallisto.tsv {output.transcripts}
         else
-        	echo "{params.exp_isl_dir}/transcripts.{wildcards.metric}.kallisto.tsv not found - skipping"
+            echo "$expIslDir/transcripts.tpm.kallisto.tsv not found - skipping"
+            exit 1
```
pcm32: if we exit here this will be an error. I understand that some experiments (ie. controlled access or manually generated stuff) are allowed not to have transcripts. Or are you controlling which experiments will skip this rule elsewhere? If keeping the exit, I would remove the skipping part and just leave snakemake to fail on the missing file when copying.

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+        set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
+        exec &> "{log}"
+
+        {workflow.basedir}/bin/rnaseqQC.sh {wildcards.accession} {workflow.basedir}
         qcExitCode=$?
 
         if [ "$qcExitCode" -eq 2 ]; then
-        	echo "Experiment $expAcc has been disqualified due to insufficient quality, exiting"
+            echo "Experiment {wildcards.accession} has been disqualified due to insufficient quality, exiting"
             exit 0
```
pcm32: I wonder here we should exit with an error or not.... otherwise we might think that the experiment got reprocessed, when in fact it didn't, and we will try to reload then.

pcm32: I would exit with an error here better.

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+
+rule quantile_normalise_expression:
+    """
+    Quantile normalize and summarize expression in tpms and fpkms.
+    To maintain previous logic, the output file is marked as temporal
+    """
+    conda: "envs/quantile.yaml"
+    log: "logs/{accession}-quantile_normalise_expression_{metric}.log"
+    input:
+        xml="{accession}-configuration.xml",
```
pcm32: given that the configuration xml is being used here explicitly, I don't think that you need the rule upstream that asks for it to be present.

pcm32: I would also name this input `config_xml` or `config`, the format is rather the less relevant bit here.

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+
+        perl {workflow.basedir}/bin/gxa_summarize_expression.pl  \
+            --aggregate-quartiles \
+            --configuration {input.xml}  \
+            < {input.qn_expression}  \
+            > {output.sum_expression}
+        if [ $? -ne 0 ]; then
+	        echo "ERROR: Failed to summarize gene expression in {wildcards.metric} for {wildcards.accession} " >&2
+	        exit 1
+        fi
```
```suggestion
```
I would remove this bit.

pcm32: you probably won't get there anyway I think.

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+        echo "ISL dir: $expIslDir"
+
+        [ ! -z $expIslDir+x ] || (echo "snakemake param exp_isl_dir needs to defined in rule" && exit 1)
+
+        if [ -s "$expIslDir/transcripts.raw.kallisto.tsv" ] ; then
+            {workflow.basedir}/bin/transcripts_expr_values_check.R {input.transcripts} $expIslDir/transcripts.raw.kallisto.tsv
+            echo "transcripts NA check -  executed for {input.transcripts} "
+        else
+            echo "$expIslDir/transcripts.raw.kallisto.tsv not found for {wildcards.accession} - skipping rule_transcripts_na_check for {input.transcripts}"
+            #exit 1
```
```suggestion
```

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+        exec &> "{log}"
+
+        if [ -s {params.transcripts} ] ; then
+
+            {workflow.basedir}/bin/quantile_normalize.sh -c {input.xml} -s {params.transcripts} -d {params.qntranscripts} -b {workflow.basedir}/bin
+
+            if [ $? -ne 0 ]; then
+                echo "ERROR: Failed to quantile normalize for {params.transcripts}  " >&2
+                exit 1
+            fi
```
```suggestion
```

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+
+            perl {workflow.basedir}/bin/gxa_summarize_expression.pl  \
+                --aggregate-quartiles \
+                --configuration {input.xml}  \
+                < {params.qn_transcripts}  \
+                > {params.agg_transcripts}
+            if [ $? -ne 0 ]; then
+                echo "ERROR: Failed to summarize transcript expressions TPMs for {params.qn_transcripts}  " >&2
+                exit 1
+            fi
```
```suggestion
```

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+            # maintain previous behaviour - quantile_normalized files are temporary
+            rm {params.qn_transcripts}
+        else
+            echo "File {params.qn_transcripts} not found. Transcript summary not performed "
+        fi
+        touch {output}
+        """
+
+
+rule generate_methods_baseline:
```
pcm32: Maybe we should be explicit in all rule names if they are rna_seq specific, as we will soon need to add the proteomics ones. In this case, this one is clearly different proteomics to rna_seq (sources of these methods files will differ).

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+        echo $de_quantMethod
+        echo $de_deMethod
+
+        perl {workflow.basedir}/bin/gxa_generate_methods.pl "$expIslDir/irap.versions.tsv" {wildcards.accession} {params.organism} {params.template} "${{baseline_mapper:?}}" "${{baseline_quantMethod:?}}" "${{de_mapper:?}}" "${{de_
+quantMethod:?}}" "${{de_deMethod:?}}" > {output.methods}
+
+        if [ $? -ne 0 ]; then
+            echo "ERROR: Failed to generate analysis methods for {wildcards.accession}" >&2
+            exit 1
+        fi
```
pcm32, line 1071:
```suggestion
```

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+        amm -s {workflow.basedir}/bin/decorateFile.sc \
+        --geneNameFile "$geneNameFile" \
+        --source {input.expression} \
+        | awk 'NR == 1; NR > 1 {{print $0 | "sort -n"}}' \
+        > $decoratedFile.swp
+
+        decoratedFileLength=$(wc -l "$decoratedFile.swp" | cut -f 1 -d ' ' )
+        if [ -s "$decoratedFile.swp" ] && [ "$decoratedFileLength" -gt 1 ]; then
+            mv $decoratedFile.swp $decoratedFile
+        else
```
pcm32, line 1114: I suspect that this won't be reached, I would remove it.

Snakefile
```diff
@@ -440,72 +513,452 @@ rule copy_raw_gene_counts_from_isl:
...
+            rm {params.output} 
+        fi
+
+        ln -s {input.bedGraph} {params.output}
+
+        if [ $? -ne 0 ]; then
+          echo "ERROR: creating symkink for {wildcards.accession} {wildcards.metric} {wildcards.assay_id}"
+          exit 1
+        fi
+
```
```suggestion
```

pcm32 (commented):
Some initial thoughts and suggestions to improve this. Good work on getting this to work  @pmb59 !

Snakefile
pcm32: Apologies, this is wrong, at it gives a missing key error if done!! Please ignore!

Snakefile
pmb59: Recalculations metrics is based on the availability of previously computed `{accession}-{metric}.tsv`, while for new experiment processing other logic should be implemented to get the metrics

Snakefile
pmb59: Done d16ae7f

Snakefile
pmb59: Done d16ae7f

Snakefile
pmb59: done 6510ad0

Snakefile
pmb59: You are right, current logic should not break if file does not exist.

Snakefile
pmb59: I agree, thanks

Snakefile
pmb59: agree

Snakefile
pmb59: thanks @pcm32 why do you think `perl gxa_summarize_expression.pl` won't produce errors?

Snakefile
pmb59: idem why do you think ` quantile_normalize.sh` won't produce errors?

Snakefile
pmb59: agree, renamed to `rule generate_methods_baseline_rnaseq`

Snakefile
pmb59: happy to remove this one!

Snakefile
pmb59: Done, thanks!

bin/round_log2_fold_changes.R
```diff
@@ -0,0 +1,48 @@
...
+# Script to round log2 fold-changes from differential expression analysis to
+# the nearest 0.1.
+
+suppressPackageStartupMessages(library(data.table))
+
+# Get commandline arguments.
+args <- commandArgs( TRUE )
+
+if( length( args ) != 3 ) {
+  stop( "\nUsage:\n\tround_log2_fold_changes.R <experiment_type> <unrounded_analytics_filename> <rounded_analytics_filename> \n\n" )
```
pcm32: When using too many positional arguments please use R optparse instead with flags.

bin/round_log2_fold_changes.R
```diff
@@ -0,0 +1,48 @@
...
+		col_l <- grep(pattern=paste0(i,'.log2foldchange'), x=colnames(unrounded_dt), fixed = TRUE )  
+		if ( length(col_l) == 0 ) { col_l <- grep(pattern=paste0(i,'.foldChange'), x=colnames(unrounded_dt), fixed = TRUE)  } 
+		# swap columns if necessary
+		if( col_l < col_p){ reordering <- replace(reordering, c(col_l, col_p), reordering[c(col_p, col_l)]) } 
+	}
+	# reorder
+	setcolorder(unrounded_dt, reordering )
+} 
+# Write out
+write.table( unrounded_dt, file = rounded, row.names=FALSE, quote=FALSE, sep="\t" )
```
pcm32, line 60: Are we 100% sure that the output here is the same in every formatting respect as generated by the previous perl script? Might be good to run this against a couple of different experiments for each technology and type to check. Otherwise something else down the line will break.

pcm32 (commented):
Looks good in general, besides optparse, please just make really sure that the output is entirely equivalent in terms of formatting to the previous one (this review is only about this round_log2_fold script).

bin/round_log2_fold_changes.R
pmb59: I noticed the perl script was forcing the `NAs` to be `0` after rounding. I will replicate that behaviour for consistency. Otherwise, same formatting for differential rna-seq and microarrays (not so well tested for proteomics,  but I will)

bin/round_log2_fold_changes.R
pmb59: done, thanks

Snakefile-reprocess
```diff
@@ -1,26 +1,170 @@
...
-
+include: "Snakefile-proteomics"
+  
 # atom: set grammar=python:
 
+def get_array_design_from_xml():
+    """
+    Parse xml config file here.
+    """
+    from xml.dom import minidom
```
pcm32, line 10: In general I would avoid DOM based approaches for parsing XML as they load the entire object in memory, but if these XML files are not that big for microarrays, this should be fine.

bin/reprocessing_routines.sh
```diff
@@ -207,3 +207,12 @@ applyFixes() {
 	fi
     done
 }
+
+# Returns prod or test, depending on the Atlas environment in which the script calling it is running
+# It is assuming that all atlasinstall_<env>s are under ${ATLAS_PROD}/sw (it will fail otherwise)
+atlas_env() {
```
pinin4fjords, line 217: Sorry, should have made the suggestion earlier. But if you're moving away from atlas-prod maybe functions like this could be called from atlas-bash-util? https://github.com/ebi-gene-expression-group/atlas-bash-util/blob/2d9c0893de2e1fe98c67f320f74466c2d2351646/generic_routines.sh#L71

With atlas-bash-util installed this call would become:

```
atlas-bash-util atlas_env
```

pmb59: thanks @pinin4fjords, this will be addressed globally in a different PR as indicated in the description.

envs/differential-stats.yaml
```diff
@@ -6,6 +6,7 @@ dependencies:
  - perl-atlas-modules >=0.3.1
  - r-atlas-internal>=1.8
  - r-ggplot2
+ - r-reshape2
```
pinin4fjords, line 9: Just a note for as and when this starts being used with containers- there is already a container for this built from https://github.com/ebi-gene-expression-group/atlas-containers/tree/master/atlas-analysis-base

Snakefile-sorting-hat
```diff
@@ -88,7 +88,41 @@ def get_check_sp_file():
...
+    if 'methods_dif' in config:
+        return config['methods_dif']
+    else:
+        return None
 
+def get_isl_dir():
+    if 'isl_dir' in config:
+        return config['isl_dir']
+    else:
+        return None
```
pcm32: Will the calling methods of these getters survive if the config is not set? If there is no other case, you might as well leave the direct config call and let snakemake fail directly. I would use these methods when there is a reasonable default for when config is not set, or if the rule actually continues to make sense when they are not defined.

pcm32 (commented):
LGTM in general. I did skip a lot of scripts because I suspect that most of them are simply copies from atlas-prod and that is not what we need to be reviewing here. But it would be good if you add a comment (in a review manner) on every script that is completely new (not copied from Atlas), to give those a more detailed inspection. Thanks!

Snakefile
pmb59: thanks @pcm32 This check was inherited from previous code. Why do you think it's unnecessary?

bin/round_log2_fold_changes.R
pmb59: PS. Proteomics will be addressed in a different PR

envs/differential-stats.yaml
pmb59: Thanks @pinin4fjords  I have added a note. Not sure at the moment how the version restriction will `affect perl-atlas-modules` and `r-atlas-internal`, but it would be good to remove those restrictions.

pmb59:
> LGTM in general. I did skip a lot of scripts because I suspect that most of them are simply copies from atlas-prod and that is not what we need to be reviewing here. But it would be good if you add a comment (in a review manner) on every script that is completely new (not copied from Atlas), to give those a more detailed inspection. Thanks!

thanks @pcm32 in general most of atlas-prod scripts required small fixes during testing, which did not affect the overall functionality, but not simple copies. I will highlight in the description what the most important scripts and file changes are.

Snakefile-sorting-hat
pmb59: thanks, done here [bf080aa](https://github.com/ebi-gene-expression-group/bulk-recalculations/pull/24/commits/bf080aa187bd5ca73b69de483c49674d2439258e)

Snakefile
```diff
@@ -420,92 +602,1307 @@ rule atlas_experiment_summary:
...
+######################################################
+# Potential new rule, to be implemented
+
+rule delete_experiment:
+    """
+    Delete experiment from Atlas.
+    It should also remove its RUN IDs from ISL db.
+    add goal='delete' in addition to 'recalculations' and 'reprocess'
+    """
```
pmb59: @pcm32 I've noticed this is something done by the curation team under current _Updating existing experiments_ SOP. Could you confirm we should not implement this here?

pcm32: Currently this happens via a Jenkins job, which probably today is using a call to the web application besides some deletions... I'm not sure how we are doing and to what extent is done (solr, db, file system), but we will need a system for this. I suspect that currently the easiest is to continue to rely on the web application call until we devise a web-cli call for this... or have we already?

pcm32: ahhh, ok, no implementation here yet. I would leave it (as in not doing it - remove it if you want) in the meantime. The Jenkins job is surely relying on the web application for this.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "/Users/pmadrigal/bulk-recalculations/isl"]
```
pcm32: This submodule should be added in a generic manner (relative paths and not bound to a user).

pcm32: In this format, it won't work on other user's machines I expect.

Snakefile
```diff
@@ -94,6 +142,51 @@ def get_metrics():
...
+        acc = config['accession']
+        files_ils_dir = os.listdir( f"{isl_dir}/{acc}/{organism}" )
+        # we only need one match, no need to traverse the full list
+        if next((s for s in files_ils_dir if '.tpm.' in s), None):
+            metrics_reprocess.append('tpms')
+        if next((s for s in files_ils_dir if '.fpkm.' in s), None):
+            metrics_reprocess.append('fpkms')
+
+        if not metrics_reprocess:
+            sys.exit("No metrics for reprocessing found in isl path.")
```
pcm32, line 165: Is this exiting with a non zero status to show failure? can we set it like that otherwise? Thanks.

Snakefile
```diff
@@ -94,6 +142,51 @@ def get_metrics():
...
+    """
+    When goal is reprocess, return config parameters to establish connection with isl db.
+    """
+    isl_vars=[ 'oracle_home', 'python_user', 'python_connect_string', 'python_password' ] 
+    db_params = []
+    for i in isl_vars:
+        if i in config:
+            db_params.append( config[i] )
+        else:
+            print(f" Missing db param: {i}")
```
```suggestion
            print(f" Missing ISL db param: {i}")
```

Snakefile
```diff
@@ -94,6 +142,51 @@ def get_metrics():
...
+            #sys.exit()
```
```suggestion
            sys.exit(1)
```

Snakefile
```diff
@@ -164,18 +256,113 @@ def get_mem_mb(wildcards, attempt):
...
     return mem_avail[attempt-1] * 1000
 
+def input_percentile_ranks(wildcards):
+    """
+    Return appropriate input for experiment type and analysis goal
+    for rule percentile_ranks.
+    """
+    if config['goal'] == 'reprocess':
+        if experiment_type=='rnaseq_mrna_differential':
+            return [ 'logs/'+wildcards['accession']+'-decorate_differential_rnaseq.done' ]
```
```suggestion
            return [ f"logs/{wildcards['accession']}-decorate_differential_rnaseq.done" ]
```
prefer f strings in general

Snakefile
```diff
@@ -164,18 +256,113 @@ def get_mem_mb(wildcards, attempt):
...
+        elif experiment_type == 'microarray_1colour_mrna_differential' or experiment_type =='microarray_2colour_mrna_differential' or experiment_type =='microarray_1colour_microrna_differential':
+            inputs = []
+            arr_designs=get_array_design_from_xml()
+            for s in arr_designs:
+                inputs.append( 'logs/'+wildcards['accession']+'_'+s+'-decorate_differential_microarray.done' )
```
pcm32: f strings here please as well.

Snakefile
```diff
@@ -164,18 +256,113 @@ def get_mem_mb(wildcards, attempt):
...
+            return inputs
+        else:
+            return None
+    if config['goal'] == 'recalculations':
+        # No input file needed - trick to force rule execution
+        return wildcards['accession']+'.metadata_summary.yaml'
```
```suggestion
        return f"{wildcards['accession']}.metadata_summary.yaml"
```

Snakefile
```diff
@@ -164,18 +256,113 @@ def get_mem_mb(wildcards, attempt):
...
+
+def input_differential_tracks_and_gsea(wildcards):
+    """
+    Return appropriate input for experiment type and analysis goal
+    for rule differential_tracks and rule differential_gsea
+    """
+    if config['goal'] == 'reprocess':
+        if experiment_type=='rnaseq_mrna_differential':
+            return [ 'logs/'+wildcards['accession']+'-decorate_differential_rnaseq.done' ]
```
```suggestion
            return [ f"logs/{wildcards['accession']}-decorate_differential_rnaseq.done" ]
```

Snakefile
```diff
@@ -164,18 +256,113 @@ def get_mem_mb(wildcards, attempt):
...
+        elif experiment_type == 'microarray_1colour_mrna_differential' or experiment_type =='microarray_2colour_mrna_differential' or experiment_type =='microarray_1colour_microrna_differential':
+            inputs = []
+            arr_designs=get_array_design_from_xml()
+            for s in arr_designs:
+                inputs.append( 'logs/'+wildcards['accession']+'_'+s+'-decorate_differential_microarray.done' )
```
```suggestion
                inputs.append( f"logs/{wildcards['accession']}_{s}-decorate_differential_microarray.done" )
```

Snakefile
```diff
@@ -164,18 +256,113 @@ def get_mem_mb(wildcards, attempt):
...
+    for rule atlas_experiment_summary
+    """
+    if config['goal'] == 'reprocess':
+        if experiment_type =='rnaseq_mrna_baseline' or experiment_type=='rnaseq_mrna_differential': 
+            return [ wildcards['accession']+'-raw-counts.tsv.undecorated' ]
+        elif experiment_type == 'microarray_1colour_mrna_differential' or experiment_type =='microarray_1colour_microrna_differential':
+            inputs = []
+            arr_designs=get_array_design_from_xml()
+            for s in arr_designs:
+                inputs.append( wildcards['accession']+'_'+s+'-normalized-expressions.tsv.undecorated' )
```
pcm32: f string

Snakefile
```diff
@@ -164,18 +256,113 @@ def get_mem_mb(wildcards, attempt):
...
+            return inputs
+        elif experiment_type =='microarray_2colour_mrna_differential':
+            inputs = []
+            arr_designs=get_array_design_from_xml()
+            for s in arr_designs:
+                inputs.append( wildcards['accession']+'_'+s+'-log-fold-changes.tsv.undecorated' )
+                inputs.append( wildcards['accession']+'_'+s+'-average-intensities.tsv.undecorated' )
```
pcm32: f string

Snakefile
```diff
@@ -164,18 +256,113 @@ def get_mem_mb(wildcards, attempt):
...
+
+
+def get_checkpoints_cp_atlas_exps(wildcards):
+    """
+    Only for reprocessing, to enable rule copy_experiment_from_analysis_to_atlas_exps
+    """
+    if config['goal'] == 'reprocess':
+        inputs = get_outputs()
+        inputs.remove('logs/'+wildcards['accession']+'-copy_experiment_from_analysis_to_atlas_exps.done')
+        inputs.remove('logs/'+wildcards['accession']+'-get_magetab_for_experiment.done')
```
pcm32: f string

Snakefile
```diff
@@ -164,18 +256,113 @@ def get_mem_mb(wildcards, attempt):
...
+        return inputs
+    else:
+        return None
 
 
-localrules: check_differential_gsea, link_baseline_coexpression, link_baseline_heatmap
+localrules: check_differential_gsea, link_baseline_coexpression, link_baseline_heatmap, copy_raw_gene_counts_from_isl, copy_normalised_counts_from_isl, copy_transcript_files_from_isl, copy_transcript_relative_isoforms, create_tracks_symlinks, check_mvaPlot_rnaseq, check_normalized_expressions_microarray, delete_intermediate_files_microarray, touch_inputs_baseline
```
pcm32: I wonder if we should leave the large copies to be sent over to the cluster? Otherwise you will have a bottle neck here which will delay in sending other jobs possibly. If copies are small, I would keep them here, but for large copies I would send them as normal jobs (not localrules).

Snakefile
```diff
@@ -377,16 +588,17 @@ rule baseline_heatmap:
...
+        expression=lambda wildcards: f"{wildcards.accession}-{wildcards.metric}.tsv" if 'reprocess' in config['goal'] else f"{wildcards.accession}-{wildcards.metric}-touch_inputs_baseline.done"
     output:
         heatmap="{accession}-heatmap-{metric}.pdf"
     shell:
         """
         set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
         exec &> "{log}"
         {workflow.basedir}/bin/generateBaselineHeatmap.R --configuration {wildcards.accession}-configuration.xml \
-		--input  {input.expression} \
+		--input  {params.expression} \
```
pcm32: This looks a bit weird, we have expression in both params and input. If the expression is a file (or set of files), we should use input, as being in params snakemake doesn't make the typical checks done for a file (input is always a file unless tagged as a directory I think).

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        export PYTHON_CONNECT_STRING={params.db_params[2]}
+        export PYTHON_PASSWORD={params.db_params[3]}
+
+        python {workflow.basedir}/isl/db/scripts/get_run_ids_atlas_prod.py {input.config_xml}
+        if [ $? -ne 0 ]; then
+	        echo "ERROR: Failed to parse atlas config file and get run ids for {wildcards.accession} " >&2
+	        exit 1
+        fi
+        touch {output}
+        """
```
pcm32, line 675: I wonder if this rule should be part of reprocess/recalculations? This seems like an undesirable circular loop between our processes. What do we need this for currently?

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        echo "ISL dir: $expIslDir"
+
+        [ ! -z $expIslDir+x ] || (echo "snakemake param exp_isl_dir needs to defined in rule" && exit 1)
+
+        if [ -s "$expIslDir/transcripts.tpm.kallisto.tsv" ] ; then
+            rsync -avz $expIslDir/transcripts.tpm.kallisto.tsv {output.transcripts}
         else
-        	echo "{params.exp_isl_dir}/transcripts.{wildcards.metric}.kallisto.tsv not found - skipping"
+            echo "$expIslDir/transcripts.tpm.kallisto.tsv not found - skipping"
         fi
```
pcm32, line 769: Why is this explicit on tpm and not parametrised by `{metric}`?

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
+        exec &> "{log}"
+        expIslDir={params.isl_dir}/{wildcards.accession}/{params.organism}
+        echo "ISL dir: $expIslDir"
+
+        [ ! -z $expIslDir+x ] || (echo "snakemake param exp_isl_dir needs to defined in rule" && exit 1)
+
         if [ -s "$expIslDir/transcripts.riu.kallisto.tsv" ] ; then
-            cp $expIslDir/transcripts.riu.kallisto.tsv {output.transcripts_relative_isoforms}
+            rsync -avz $expIslDir/transcripts.riu.kallisto.tsv {params.transcripts_relative_isoforms}
```
pcm32, line 794: This pattern seems a bit hacky, couldn't we rely on the presence / absence of the input `$expIslDir/transcripts.riu.kallisto.tsv` to decide the output?

Snakefile
pcm32: Won't `set -e` mean that it will stop and exit at the first failure? If that is the case, then you will never reach that if.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+rule transcripts_na_check:
+    """
+    Replace NAs with 0 in Kallisto TPM transcripts, if the file exists
+    (input and output file here is the same)
+    """
+    conda: "envs/atlas-internal.yaml"
+    log: "logs/{accession}-rule-transcripts_na_check_{metric}.log"
+    params:
+        organism=get_organism(),
+        isl_dir=get_isl_dir()
```
pcm32, line 887: Shouldn't this be a directory input? I think you can do something like:
```
   input:
       isl_dir=directory(get_isl_dir())
```
and then snakemake will make sure that this exists as an input, I think. Unless that this is undesirable.

Snakefile
pcm32: `set -e`

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+    log: "logs/{accession}-summarize_transcripts_{metric}.log"
+    input:
+        xml="{accession}-configuration.xml",
+        getqn=rules.quantile_normalise_transcripts.output
+    output:
+        temp("logs/{accession}-summarize_transcripts_{metric}.done")
+    params:
+        qn_transcripts="{accession}-transcripts-{metric}.tsv.undecorated.quantile_normalized",
+        agg_transcripts="{accession}-transcripts-{metric}.tsv.undecorated.aggregated"
+    shell:
```
pcm32, line 956: I suspect that it is a bit hacky again to use files as parameters like this, but I see why you do it. What is the "Snakemaky" way of doing this @pinin4fjords ?

Snakefile
pcm32: `set -e` won't let you reach this.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        if [ ! -s {params.irap_versions} ] ; then
+            echo "it does not - attempting to transfer the file from the singularity container {params.irap_container}"
+            singularity exec {params.irap_container} cp /opt/irap/aux/mk/irap_versions.mk {params.irap_versions}
+            if [ ! -s {params.irap_versions} ] ; then
+                echo "ERROR: Failed to retrieve {params.irap_versions} from singularity" >&2
+                exit 1
+            fi
+        else
+            echo "The file {params.irap_versions} already exists."
+        fi
```
pcm32: Maybe better to have a rule for this and use the resulting {irap_versions} explicitly as an input (from a central location)? It will only run once as well.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
+        exec &> "{log}"
+        source {workflow.basedir}/bin/reprocessing_routines.sh
+
+        geneNameFile=$( get_geneNameFile_given_organism {params.organism}  )
+
+        echo $geneNameFile
+
+        test -s "$geneNameFile" || (  >&2 echo "$0 gene name file not found: $geneNameFile" ; exit 1 )
+        #test -s "$FPKMexpressionsFile" -o -s "$TPMexpressionsFile" || (  >&2 echo "$0 no data files for $e" ; exit 1 )
```
```suggestion
```
if we are sure that we don't want this anymore.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+
+        decoratedFile={output.decoexpression} 
+
+        # Ammonite REPL & Script-Runner
+
+        amm -s {workflow.basedir}/bin/decorateFile.sc \
+        --geneNameFile "$geneNameFile" \
+        --source {input.expression} \
+        | awk 'NR == 1; NR > 1 {{print $0 | "sort -n"}}' \
+        > $decoratedFile.swp
```
```suggestion
            --geneNameFile "$geneNameFile" \
            --source {input.expression} \
            | awk 'NR == 1; NR > 1 {{print $0 | "sort -n"}}' \
            > $decoratedFile.swp
```

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+            echo $geneNameFile
+            echo $transcriptFile
+
+            test -s "$geneNameFile" || (  >&2 echo "$0 gene name file not found: $geneNameFile" ; exit 1 )
+            test -s "$transcriptFile" || (  >&2 echo "$0 transcript to gene mapping file not found: $transcriptFile" ; exit 1 )
+
+            decoratedFile={params.decotranscripts} 
+
+            # Ammonite REPL & Script-Runner
+            #export JAVA_OPTS="-Xmx3000M"
```
pcm32: I remember that some files required more memory for decoration, was this commented in the code?

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+                --source {params.transcripts} \
+                | awk 'NR == 1; NR > 1 {{print $0 | "sort -n"}}' \
+                > $decoratedFile.swp
+
+            decoratedFileLength=$(wc -l "$decoratedFile.swp" | cut -f 1 -d ' ' )
+            if [ -s "$decoratedFile.swp" ] && [ "$decoratedFileLength" -gt 1 ]; then
+                mv $decoratedFile.swp $decoratedFile
+            else
+                echo "ERROR: decorate_transcript_baseline_file for {wildcards.accession} and {wildcards.metric}"
+	            exit 1
```
pcm32, line 1173:
```suggestion
                echo "ERROR: decorate_transcript_baseline_file for {wildcards.accession} and {wildcards.metric}"
	        exit 1
```

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
+        exec &> "{log}"
+
+        # remove if any old bedgraph files
+        if [ -s {params.output} ] ; then
+            rm {params.output} 
+        fi
+
+        ln -s {input.bedGraph} {params.output}
+        touch {output}
```
pcm32: why not?

```suggestion
    output: "{accession}.{assay_id}.genes.expressions.bedGraph"
    shell:
        """
        set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
        exec &> "{log}"

        # remove if any old bedgraph files
        rm -f {output} 

        ln -s {input.bedGraph} {output}
```

and make sure that this output is added to the list of required outputs for the all rule, and remove the .done ?

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        if [ ! -d "$TMPDIR"/tmp ]; then
+            mkdir $TMPDIR/tmp
+        fi
+
+        perl {workflow.basedir}/bin/diffAtlas_DE.pl --experiment {wildcards.accession} --directory ./
+        if [ $? -ne 0 ]; then
+	        echo "ERROR: Failed to generate differential expression statistics and MvA plots for {wildcards.accession}" >&2
+	        rm -rf *.png {wildcards.accession}-analytics.tsv.undecorated
+	        exit 1
+        fi
```
```suggestion
```
this will never be reached due to set -e

Snakefile
```diff
@@ -16,9 +16,54 @@ def read_metadata_summary():
...
+    if 'irap_container' in config:
+        return config['irap_container']
+    else:
+        return None
+
+def get_tmp_dir():
+    if 'tmp_dir' in config:
+        return config['tmp_dir']
+    else:
+        return None
```
pcm32, line 66: I see this function used a lot, is there a need for all the tools to use the same tmp dir (so that is why it is being set from upstream) or in most case you just one independent tmp dirs (what I would guess). If that is the case, I would avoid taking this all the way to snakemake and simply use `TMPDIR=$(mktemp --directory)` on the shells (that will create a local temp directory on the machine where this is running). You need to remember to delete it of course.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        """
+        set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
+        exec &> "{log}"
+        PATH=$PATH:{workflow.basedir}/bin
+        
+        TMPDIR={params.tmp_dir}
+        echo $TMPDIR"/tmp"
+        if [ ! -d "$TMPDIR"/tmp ]; then
+            mkdir $TMPDIR/tmp
+        fi
```
pcm32: There is a lot of logic here for a temp directory, I would get rid of the params.tmp_dir and simply use:
```suggestion
        TMPDIR=$(mktemp --directory)
        echo $TMPDIR
```
is this used internally in the scripts below? I don't see it used at least at the bash level.

pcm32: the same in other occurrences of the same, if it works.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+    shell:
+        """
+        set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
+        exec &> "{log}"
+
+        if [ -s "{wildcards.accession}-{wildcards.contrast_id}-mvaPlot.png" ] ; then
+            echo "mvaPlot for {wildcards.accession} and {wildcards.contrast_id} successfully generated during differential rna-seq statistics: {wildcards.accession}-{wildcards.contrast_id}-mvaPlot.png"
+        else
+            echo "ERROR: mvaPlot for {wildcards.accession} and {wildcards.contrast_id} not produced during differential rna-seq statistics."
+            exit 1
```
pcm32, line 1256: Why not declare this file for accession/contrast_id as output of the rule that is generating them?

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        {workflow.basedir}/bin/round_log2_fold_changes.R \
+            --experiment_type {params.exp_type} \
+            --input_to_round {input} \
+            --intermediate_output {params.intermediate_rounded}  
+
+        if [ $? -ne 0 ]; then
+	        echo "ERROR: Failed to round to one decimal place log2fold changes in {input} " >&2
+	        rm -rf {params.intermediate_rounded}
+	        exit 1
+        fi
```
```suggestion
```

will not be reached.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        mv {input} {output.unrounded}
+        mv {params.intermediate_rounded} {input}
```
pcm32, line 1289: This means that this step will always be redone, is this intended? This logic is a bit strange.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+
+        echo "Cheking if IRAP versions file: {params.irap_versions} exists.."
+
+        if [ ! -s {params.irap_versions} ] ; then
+            echo "it does not - attempting to transfer the file from the singularity container {params.irap_container}"
+            singularity exec {params.irap_container} cp /opt/irap/aux/mk/irap_versions.mk {params.irap_versions}
+            if [ ! -s {params.irap_versions} ] ; then
+                echo "ERROR: Failed to retrieve {params.irap_versions} from singularity" >&2
+                exit 1
+            fi
```
pcm32: If we use the rule as I suggested before, then this wouldn't be repeated here, and you would just use the output of that rule as input here.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        do  
+            echo "${{i}}"
+            decoratedFile=`echo "${{i}}" | sed 's/\.undecorated//'`
+            echo "$decoratedFile"
+
+            amm -s {workflow.basedir}/bin/decorateFile.sc \
+            --geneNameFile "$geneNameFile" \
+            --source "${{i}}" \
+            | awk 'NR == 1; NR > 1 {{print $0 | "sort -n"}}' \
+            > $decoratedFile.swp
```
```suggestion
                --geneNameFile "$geneNameFile" \
                --source "${{i}}" \
                | awk 'NR == 1; NR > 1 {{print $0 | "sort -n"}}' \
                > $decoratedFile.swp
```

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        echo $idf_filename
+        echo $ae_dir
+        echo $mirbase_dir
+
+        # Get normalized expressions
+        perl {workflow.basedir}/bin/arrayNormalization.pl {wildcards.accession} $idf_filename $ae_dir $mirbase_dir {workflow.basedir} $(pwd)
+        if [ $? -ne 0 ]; then
+            echo "ERROR: Failed to calculate normalized expressions for {wildcards.accession}" >&2
+            exit 1
+        fi
```
```suggestion
```

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+
+        for p in {params.array_designs}
+        do  
+            echo "Array designs found in the xml config file: ${{p}}"
+            if [ -s {wildcards.accession}_"${{p}}-normalized-expressions.tsv.undecorated" ] ; then
+                echo "File {wildcards.accession}_${{p}}-normalized-expressions.tsv.undecorated exists for {wildcards.accession} and array_design ${{p}}"
+                cp {wildcards.accession}_"${{p}}-normalized-expressions.tsv.undecorated" {wildcards.accession}_"${{p}}-normalized-expressions.tsv.undecorated.unmerged"
+            else
+                echo "ERROR: File does not exist for {wildcards.accession} and array_design ${{p}}" >&2
+                exit 1
```
pcm32, line 1479: if `{wildcards.accession}_"${{p}}-normalized-expressions.tsv.undecorated"` would be an explicit output of the rule that generates it, it could be used as an input here (or a list of inputs), and reduce the amount of things here.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+
+        # check that qc dirs exist for each array design
+        for p in {params.array_designs}
+        do
+            if [ -d "./qc/{wildcards.accession}_${{p}}_QM" ]; then
+                echo " ./qc/{wildcards.accession}_${{p}}_QM found"
+            else
+                echo "ERROR: ./qc/{wildcards.accession}_${{p}}_QM not found"
+                exit 1
+            fi
```
pcm32, line 1538: again, I wonder if it is not better to declare these are outputs of the rule that generates them and consume them here are inputs. Then we don't need these checks.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+                geneNameFile=$(get_geneNameFile_given_organism {params.organism} )
+            fi
+
+            # generate temp decorated file normalized-expressions.tsv
+            decorate_if_exists_norm "${{expPath}}/{wildcards.accession}_${{arrayDesign}}-normalized-expressions.tsv.undecorated" "$arrayDesignFile" "$geneNameFile"
+        done
+        if [ $? -ne 0 ]; then
+            echo "ERROR: Generate temp decorated files for {wildcards.accession}" >&2
+            exit 1
+        fi
```
```suggestion
```

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        find ${{expPath}} -maxdepth 1 -name "{wildcards.accession}_A-*-normalized-expressions.tsv.undecorated" \
+            | xargs -n1 basename \
+            | sed "s/{wildcards.accession}_//" \
+            | sed "s/-normalized-expressions.tsv.undecorated//" \
+            | while read -r arrayDesign ; do
+            arrayDesignFile=$(get_arraydesign_file ${{arrayDesign}} {params.organism}  )
+            if [ $? -ne 0 ]; then
+                echo "ERROR: Could not find array design: $arrayDesign" >&2
+                exit 1
+            fi
```
pcm32: If we want to signal that the error is here, we could add an echo before or after the command signalling it started and/or finished.

```suggestion
```

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        if [ ! -d "$TMPDIR"/tmp ]; then
+            mkdir $TMPDIR/tmp
+        fi
+        # Calculate analytics
+        perl {workflow.basedir}/bin/diffAtlas_DE.pl --experiment {wildcards.accession} --directory ./
+        if [ $? -ne 0 ]; then
+	        echo "ERROR: Failed to calculate analytics for {wildcards.accession}" >&2
+	        rm -rf *.png *-analytics.tsv.undecorated
+	        exit 1
+        fi
```
```suggestion
```

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        """
+        set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
+        exec &> "{log}"
+        PATH=$PATH:{workflow.basedir}/bin
+        
+        TMPDIR={params.tmp_dir}
+        echo $TMPDIR"/tmp"
+        if [ ! -d "$TMPDIR"/tmp ]; then
+            mkdir $TMPDIR/tmp
+        fi
```
pcm32: lets try to reduce all of these to what I suggested, or at least replace them with a snakemake block that is used everywhere.

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+        exec &> "{log}"
+
+        for p in {params.array_designs}
+        do
+            if [ -s "{wildcards.accession}_${{p}}-analytics.tsv.undecorated" ]; then
+                {workflow.basedir}/bin/check_na_pvals.R "{wildcards.accession}_${{p}}-analytics.tsv.undecorated"
+                if [ $? -ne 0 ]; then
+                    echo "ERROR: no non-NA p-values found in {wildcards.accession}_${{p}}-analytics.tsv.undecorated" >&2
+                    exit 1
+                fi
```
```suggestion
```

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+                                                                                                                                                                                                               
+        if [ $? -ne 0 ]; then                                                                                                                                                                                  
+                echo "ERROR: Failed to round to one decimal place log2fold changes in {params.analytics} " >&2                                                                                                 
+                rm -rf {params.intermediate_rounded}                                                                                                                                                           
+                exit 1                                                                                                                                                                                         
+        fi
```
```suggestion                                                                                                                                                                                                  
```

Snakefile
```diff
@@ -420,92 +637,1344 @@ rule atlas_experiment_summary:
...
+            decorate_if_exists "${{expPath}}/{wildcards.accession}_${{arrayDesign}}-analytics.tsv.undecorated" "$arrayDesignFile" "$geneNameFile"
+            decorate_if_exists "${{expPath}}/{wildcards.accession}_${{arrayDesign}}-normalized-expressions.tsv.undecorated" "$arrayDesignFile" "$geneNameFile"
+            decorate_if_exists "${{expPath}}/{wildcards.accession}_${{arrayDesign}}-average-intensities.tsv.undecorated" "$arrayDesignFile" "$geneNameFile"
+            decorate_if_exists "${{expPath}}/{wildcards.accession}_${{arrayDesign}}-log-fold-changes.tsv.undecorated" "$arrayDesignFile" "$geneNameFile"
+            decorate_if_exists "${{expPath}}/{wildcards.accession}_${{arrayDesign}}-analytics.tsv.undecorated.unrounded" "$arrayDesignFile" "$geneNameFile"
+        done
+        if [ $? -ne 0 ]; then
+            echo "ERROR: Failed to decorate {wildcards.accession} and array {wildcards.array_design}" >&2
+            exit 1
+        fi
```
```suggestion
```

Snakefile-reprocess
```diff
@@ -1,26 +1,175 @@
...
+        outputs.append(f"{config['accession']}-coexpressions.tsv.gz")
+        #metric_link_coexp=False
+        #for m in metrics:
+        #    expression_file=f"{config['accession']}-{m}.tsv"
+        #    print(f"Checking file size for {expression_file} and number of columns")
+        #    if os.path.getsize(expression_file) > 0 and get_number_columns(expression_file)>4:
+        #        metric_link_coexp=True
+        #        outputs.extend(expand(f"{config['accession']}-{m}-coexpressions.tsv.gz"))
+        #if metric_link_coexp == True:
+        #    outputs.extend(expand(f"{config['accession']}-coexpressions.tsv.gz" ))
```
pcm32:
```suggestion
```

pcm32: if we don't want it I mean, things are fairly long already.

bin/arrayNormalization.R
```diff
@@ -0,0 +1,391 @@
...
+	
+			agilentArray( files, mode, assayNames, outFile, miRBaseFile )
+
+		}
+		# Illumina data 
+		# use lumi package for pre-processing of raw text data.
+		else if(mode == "lumi") {
+
+			illuminaArray( files, outFile )
+
```
pcm32:
```suggestion
			illuminaArray( files, outFile )
```

bin/arrayNormalization.R
```diff
@@ -0,0 +1,391 @@
...
+		# pre-processing with limma. This consists of subsetting probes based
+		# on miRBase mappings, background correction, within- or between-array
+		# normalization, and averaging over duplicates.
+		# For 1-colour data, normalized intensities are written to a single file.
+		# For 2-colour data, log2(fols change)s are written to one file and
+		# average intensities to another file.
+		if(mode == "agil1" | mode == "agil2") {
+	
+			agilentArray( files, mode, assayNames, outFile, miRBaseFile )
+
```
pcm32:
```suggestion
			agilentArray( files, mode, assayNames, outFile, miRBaseFile )
```

bin/arrayNormalization.R
```diff
@@ -0,0 +1,391 @@
...
+            
+        }
+		# For Affymetrix data, use oligo package for pre-processing: this is
+		# background adjustment, normalization and probe set summarization (see
+		# below for details).
+		else if(mode == "affy") {
+
+			affymetrixArray( files, assayNames, outFile )
+
+		}
```
pcm32:
```suggestion
		else if(mode == "affy") {
			affymetrixArray( files, assayNames, outFile )
		}
```

bin/arrayNormalization.R
```diff
@@ -0,0 +1,391 @@
...
+
+		print("Reading Agilent 1-colour data")
+		# for 1-colour Agilent data we only have a green channel.
+		dataSet <- try({read.maimages(files, source="agilent", green.only=TRUE)})
+	}
+	else if(mode == "agil2") {
+		
+		print("Reading Agilent 2-colour data")
+		dataSet <- try({read.maimages(files, source="agilent")})
+	}
```
pcm32:
```suggestion
	if(mode == "agil1") {
		print("Reading Agilent 1-colour data")
		# for 1-colour Agilent data we only have a green channel.
		dataSet <- try({read.maimages(files, source="agilent", green.only=TRUE)})
	}
	else if(mode == "agil2") {
		print("Reading Agilent 2-colour data")
		dataSet <- try({read.maimages(files, source="agilent")})
	}
```

bin/arrayNormalization.R
```diff
@@ -0,0 +1,391 @@
...
+	# NB: this might get masked by oligo package's backgroundCorrect
+	# method if they are both loaded (here they shouldn't be so don't
+	# call limma's explicitly).
+	print("Background correcting")
+	dataSet <- backgroundCorrect(dataSet, method="normexp", offset=50)
+
+	if(mode == "agil1") {
+		
+		# Between-array normalization for 1-colour.
+		normData <- normalizeBetweenArrays(dataSet, method="quantile")
```
pcm32:
```suggestion
		# Between-array normalization for 1-colour.
		normData <- normalizeBetweenArrays(dataSet, method="quantile")
```

bin/arrayQC.R
```diff
@@ -0,0 +1,312 @@
...
+	# for two-colour ones.  The values in this column are used to colour points
+	# in the plots in the report.
+	factorColName <- "FactorValue"
+
+	# Read in raw data and add factor values to object created.
+	# Affymetrix data
+	if(exptType == "affy") {
+   
+ 		dataSet <- affymetrixQC( annotationFile )
+	}
```
pcm32:
```suggestion
	if(exptType == "affy") {
 		dataSet <- affymetrixQC( annotationFile )
	}
```

bin/arrayQC.R
```diff
@@ -0,0 +1,312 @@
...
+		dataSet <- illuminaQC( annotationFile )
+
+	}
+	# Agilent data.
+	# One-colour or Two color Agilent data
+	else if(exptType == "agil1" | exptType == "agil2") {
+
+		dataSet <- agilentQC(annotationFile, exptType, miRBaseFile)
+
+	}
```
pcm32:
```suggestion
	else if(exptType == "lumi") {
		dataSet <- illuminaQC( annotationFile )
	}
	# Agilent data.
	# One-colour or Two color Agilent data
	else if(exptType == "agil1" | exptType == "agil2") {
		dataSet <- agilentQC(annotationFile, exptType, miRBaseFile)
	}
```

bin/arrayQC.R
```diff
@@ -0,0 +1,312 @@
...
+    ma = aqm.maplot(preparedData)
+    qm = list("Boxplot" = bo, "Density" = de, "Heatmap" = hm, "PCA" = pca, "MAplot" = ma)
+    arrayQCdata <- aqm.writereport( 
+        modules = qm, 
+        arrayTable = preparedData$pData, 
+        reporttitle = reportTitle, 
+        outdir = outDir
+    )
+    # End temporary workaround
+    ##########################
```
pcm32, line 86: Is this ours (recent) or has been there for years probably?

bin/arrayQC.R
```diff
@@ -0,0 +1,312 @@
...
+		# If we have a miRBase mapping file, subset the data to only include
+		# probes that are there.
+		if(miRBaseFile != 0) {
+			dataSet <- subsetProbes(dataSet, miRBaseFile)
+		}
+	}
+
+	return(dataSet)
+
+}
```
pcm32:
```suggestion
	}
	return(dataSet)
}
```

bin/check_na_pvals.R
```diff
@@ -0,0 +1,18 @@
...
+analyticsFilename <- args[ 1 ]
+
+analyticsResults <- read.delim( analyticsFilename, header=TRUE )
+
+pvalueColnames <- colnames( analyticsResults )[ grep( "p.value", colnames( analyticsResults ) ) ]
+
+nonNAcounts <- sapply( pvalueColnames, function( x ) {
+    
+    length( which( !is.na( analyticsResults[[ x ]] ) ) )
+} )
```
pcm32:
```suggestion
args <- commandArgs( TRUE )
analyticsFilename <- args[ 1 ]
analyticsResults <- read.delim( analyticsFilename, header=TRUE )
pvalueColnames <- colnames( analyticsResults )[ grep( "p.value", colnames( analyticsResults ) ) ]
nonNAcounts <- sapply( pvalueColnames, function( x ) {
    length( which( !is.na( analyticsResults[[ x ]] ) ) )
} )
```

envs/differential-stats.yaml
```diff
@@ -0,0 +1,14 @@
+channels:
```
pcm32, line 1: Maybe, now that this is well established, we should aim to pin most versions to the current ones in the working environments. But we can maybe do that on a separate PR.

pcm32: Or at least avoid that they go beyond the next major version.

run_sorting_hat_test_data.sh
```diff
@@ -21,6 +56,13 @@ CONDA_PREFIX_LINE="--conda-prefix $SN_CONDA_PREFIX"
 export LOG_PATH=${LOG_PATH:-$( pwd )/sorting.log}
 USUAL_SM_ERR_OUT=${USUAL_SM_ERR_OUT:-$( pwd )/snakemake.log}
 
+# isl db
+ORACLE_BASE=
+ORACLE_HOME=
+PYTHON_USER=
+PYTHON_CONNECT_STRING=
+PYTHON_PASSWORD=
```
pcm32: If this is meant to be like a sample script and the one with all the real values is stored elsewhere, maybe we can then rename this to `*.sample.sh`.

pcm32: Otherwise I would suggest that this script can read these vars from outside and then complain if not set.

pcm32 (commented):
A titanic effort @pmb59, well done. Looks very good in general, I only have some minor concerns for some patterns that might be adding more code than we need in the bash parts (ie. using files as params instead of as output inputs means that we need to be checking manually certain things that snakemake will check for you by default). But if those are major changes we could as well open issues and fix those iteratively later, specially once we have tests for this whole thing. There is a lot of lines that we can get rid of due to the `set -e` not allowing us to reach certain inherited checks, in some cases it might be useful to signal where the process ended though for the operator. We can discuss this on a call if it is easier.

pcm32:
Maybe we could also update the README to reflect what can this repo do after these changes.

.gitmodules
pmb59: This was changed here [7806fc0](https://github.com/ebi-gene-expression-group/bulk-recalculations/commit/7806fc051848f78b93be23e5808502479dc3b8ca)

Snakefile
pmb59: Yes. If a non-integer is passed, `None` is equivalent to passing 0, and any other object (such as the string here) is printed to stderr and results in an exit code of 1.

Snakefile
pmb59: done 1fdfe63

Snakefile
pmb59: 8ff665f

Snakefile
pmb59: 8ff665f

Snakefile
pmb59: 82ea574

Snakefile
pmb59: thanks @pcm32  that makes sense 2931bf1

Snakefile
pmb59: thanks @pcm32 I've removed the params 8ee9634

Snakefile
pmb59: As discussed today, it depends on whether the cron job deals with this task or not. In the worst case scenario, keeping the rule is harmless:
- it will insert the run_ids collected into LIBRARIES_ATLAS, if they have not been inserted before.
- otherwise it does nothing, except some qc checks. For instance if there are no valid run_ids in the experiment will raise an ERROR. In that case it is also convenient to stop running the workflow. I will also trigger WARN if there are duplicated Run ids or with unexpected accessions.

Snakefile
pmb59: Thanks @pcm32, this is an important point. This code was inherited from [here](https://github.com/ebi-gene-expression-group/atlas-prod/blob/develop/db/scripts/import_baseline_rnaseq_experiment.sh#L39-L51). Should we bring `transcripts.fpkm.kallisto.tsv` to data production? 
Pinning @pinin4fjords in case he wants to comment.

Snakefile
pmb59: TTBOMK, this keyword is only valid for outputs in Snakemake.

run_sorting_hat_test_data.sh
pmb59: thanks, I agree a859e53

Snakefile
pmb59: thanks @pcm32  The flag `   export JAVA_OPTS="-Xmx3000M"` was used in the old setup. Actually only for baseline transcript decoration [here](https://github.com/ebi-gene-expression-group/atlas-prod/blob/develop/bioentity_annotations/decorate_routines.sh#L57). Ammonite man says the following: 

> Set a default heap limit of 500 megabytes to the amm standalone executable, so it no longer grows to the "default" of 1/4 your system memory (typically multiple gigabytes) unnecessarily. If you need to give it more memory, start it with JAVA_OPTS=-Xmx1500M amm

I don't think we have experienced problems in this rule, but I will provide ammonite available memory on the workflows (which depends on Snakemake attempts), just in case

pmb59: done 971c0d4

Snakefile
pmb59: done 8e122dd

Snakefile
pcm32: I guess that if we have fpkms for transcripts we should bring it over, right? Can you check if we have them in general (I guess in noah) among the ISL results?

pcm32: In any case, if this wasn't part of the original code, I'm happy for us to leave this like this towards closing this monster PR and open an issue and then resolve it in a smaller PR.

pinin4fjords: I'm afraid that I don't know what the rationale was of our predecessors with the original code, so I don't have a useful comment to make here.

pcm32: Issue: https://github.com/ebi-gene-expression-group/bulk-recalculations/issues/32

Snakefile
pmb59: @pcm32 I did that to avoid to avoid the uncertainty error in Snakemake produced because the output file has no wildcard `{metric}` on it. Hence the required output was: 
```
        outputs.extend(expand(f"{config['accession']}"+".{a_id}.create_tracks_symlinks_{metric}.done",
                            a_id=get_assay_ids(),
                            metric=["tpms"]))
```
Now, I have removed this wildcard, as only TPMs are required, 68be696

Snakefile
pmb59: thanks - `transcripts.fpkm.kallisto.tsv` are available, but have not been reported by Atlas so far (see the FTP). I will add a new feature in a different branch to allow this.

Snakefile
pmb59: 5198675

Snakefile
pmb59: d95eb9c

Snakefile
pmb59: 5f6f770

pmb59: 5f6f770

Snakefile
pmb59: 64cfd7e

Snakefile
pmb59: 355c51c

Snakefile
pmb59: See issue https://github.com/ebi-gene-expression-group/bulk-recalculations/issues/33

Snakefile
pmb59: See https://github.com/ebi-gene-expression-group/bulk-recalculations/issues/33

Snakefile
pmb59: See issue https://github.com/ebi-gene-expression-group/bulk-recalculations/issues/33

Snakefile
pmb59: This will be handled in issue https://github.com/ebi-gene-expression-group/bulk-recalculations/issues/33

Snakefile
pmb59: This will be handled in issue https://github.com/ebi-gene-expression-group/bulk-recalculations/issues/33

envs/differential-stats.yaml
pmb59: See issue https://github.com/ebi-gene-expression-group/bulk-recalculations/issues/34

bin/arrayQC.R
pmb59: indeed, this chunk has been inherited from our predecessors

Snakefile
pmb59: Done 61a9fd1. This logic was necessary to avoid:

> Not all output, log and benchmark files of rule get_irap_versions_file contain the same wildcards. This is crucial though, in order to avoid that two or more jobs write to the same file.

Snakefile
pmb59: done

Snakefile
pmb59: Unfortunately, currently mktemp won't create a temp dir name in appropriate codon path. But I have included this option in case we do not want a custom TMP_DIR 18bd548

Snakefile
pmb59: Currently we depend on a custom temp_dir location in codon (see comment above), but I have moved basic logic to the common routines script f678934

Snakefile
pmb59: Currently we depend on a custom temp_dir location in codon (see comment above), but I have moved basic logic to the common routines script f678934

Snakefile
pmb59: Fully agree, but I respected the Atlas logic defined previously [here](https://github.com/ebi-gene-expression-group/atlas-prod/blob/develop/db/scripts/import_differential_rnaseq_experiment.sh#L59-L60). 
In my experience this rule is not redone during reprocessing, at least with forceall=true, but I have left an issue to investigate further https://github.com/ebi-gene-expression-group/bulk-recalculations/issues/39

PR #26, opened 2021-11-16
add  LSF config yaml to Snakemake

pmb59:
This PR adds logic to the sorting hat to copy the lsf_config yaml to the working directory

### DIFF
--- a/Snakefile-sorting-hat
+++ b/Snakefile-sorting-hat
@@ -252,8 +252,9 @@ rule produce_recalculations_call:
         mkdir -p {wildcards.accession}/logs
         exec &> {log}
         pushd {input.working_directory}
-        cp {params.lsf_config} lsf.yaml
-
+	if ! [ -z {params.lsf_config} ]; then 
+            ln -s {params.lsf_config} lsf.yaml
+	fi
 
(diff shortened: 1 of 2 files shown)


Snakefile-sorting-hat
```diff
@@ -219,7 +219,7 @@ rule produce_reprocess_call:
         mkdir -p {wildcards.accession}/logs
         exec &> {log}
         pushd {input.working_directory}
-        cp {params.lsf_config} {input.working_directory}
+        cp {params.lsf_config} {wildcards.accession}/lsf.yaml
```
pcm32: I would make this optional, so that it happens only if that params.lsf_config is defined.

pcm32: Why not just link the file, then if you need to edit hosts in the middle of the run it can be done on a single place?

pcm32 (approved):
A few minor suggestions. Feel free to merge after considering them.

Snakefile-sorting-hat
pmb59: @pcm32  I have implemented both suggestions. Tested and it works fine

PR #28, opened 2021-11-18
Feature/moremem to percentile

pmb59:
This PR: 
- Increases memory in percentile_ranks with each attempt
- Removes `lsf.yaml` if present in an accession study to avoid error before creating the symlink

### DIFF
--- a/Snakefile-sorting-hat
+++ b/Snakefile-sorting-hat
@@ -253,10 +253,9 @@ rule produce_recalculations_call:
         exec &> {log}
         pushd {input.working_directory}
         if ! [ -z {params.lsf_config} ]; then
-            if [ -f "lsf.yaml" ]; then
-                rm -f lsf.yaml
+            if [ ! -f "lsf.yaml" ]; then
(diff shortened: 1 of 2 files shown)


Snakefile-sorting-hat
```diff
@@ -252,9 +252,12 @@ rule produce_recalculations_call:
         mkdir -p {wildcards.accession}/logs
         exec &> {log}
         pushd {input.working_directory}
-	if ! [ -z {params.lsf_config} ]; then 
+        if ! [ -z {params.lsf_config} ]; then
+            if [ -f "lsf.yaml" ]; then
+                rm -f lsf.yaml
+            fi
             ln -s {params.lsf_config} lsf.yaml
```
```suggestion
            if [ ! -f "lsf.yaml" ]; then
                ln -s {params.lsf_config} lsf.yaml
            fi
```
might be less expensive to do this instead of deleting and creating every time.

pcm32 (approved):
LGTM, I have added a small suggestion that could reduce operations done.

PR #29, opened 2021-11-18
New logic for memory assignment after each attempt

pmb59:
This PR changes the logic on memory assignment with different reiterations, based on a predefined set of values


pcm32 (approved):
LGTM

PR #30, opened 2021-11-19
Adds the ability to set the strict flag of the assay factory

pcm32:
Some experiments fail to obtain metadata due to empty factor values for assays that are not part of Atlas, in an over zealous strictness at a level where we only want to get magetab paths and organisms.

- [x] Adds option to perl script
- [x] Use option in sorting hat


pmb59 (approved):
LGTM!

PR #31, opened 2021-12-10
+1 correction for the log of expression values for the coexpression matrix

pmb59:
When creating a gene coexpression matrix for a baseline Expression Atlas experiment certain checks are applied, such as making sure the values are numeric or converting any NAs to 0. When expression values are 0, log will produce `-Inf` in R. With the simple correction proposed here we aim to avoid that.


pcm32 (approved):
LGTM, great that you resolved this!

PR #40, opened 2022-05-24
Refactor bash functions to use atlas-bash-util

pmb59:
In the repo there are functions that are already available here:
https://github.com/ebi-gene-expression-group/atlas-bash-util

The goal is to do some code refactoring to make use of atlas-bash-util. 

A submodule, rather than activating a new conda env in multiple rules, seems fine as mostly sourcing bash functions here and the dependencies of atlas-bash-utils are ensured (bash, coreutils, grep).


Snakefile
```diff
@@ -1090,6 +1087,7 @@ rule decorate_expression_baseline:
         set -e # snakemake on the cluster doesn't stop on error when --keep-going is set
         exec &> "{log}"
         source {workflow.basedir}/bin/reprocessing_routines.sh
+        source {workflow.basedir}/atlas-bash-util/generic_routines.sh
```
pcm32, line 1090: if you are only adding a source here, how were these variables provisioned before? The ones that you are getting from bash-util/generic_routines?

Snakefile-sorting-hat
```diff
@@ -165,7 +165,7 @@ rule get_experiment_metadata:
         export ATLAS_META_CONFIG={params.atlas_meta_config}
         export ATLAS_PROD={params.atlas_prod}
         export ATLAS_EXPS={params.atlas_exps}
-        source {workflow.basedir}/bin/metadata_retrieval_functions.sh
+        source {workflow.basedir}/atlas-bash-util/generic_routines.sh
```
pcm32, line 168: My concern here is that metadata_retrieval_functions was very specific and generic routines is quite wide? Is this replacement working well?

bin/quantile_normalize.sh
```diff
@@ -44,7 +44,7 @@ trap 'rm -f ${tmp_files}.*' INT TERM EXIT
 
 tmp_config_tsv=$tmp_files.config.tsv
 #"$scriptDir/get_assay_groups.pl" "$path_to_experiment_config" > "$tmp_config_tsv"
-"$scriptDir/get_assay_groups.pl" "$path_to_experiment_config" > "$tmp_config_tsv"
+"$scriptDir/atlas-bash-util/get_assay_groups.pl" "$path_to_experiment_config" > "$tmp_config_tsv"
```
pcm32, line 47: Are these identical? I remember having to add some changes to the version I brought to this repo. Or maybe it was another function.

pcm32 (approved):
Looks good in general, really biggest thing here is whether this all works when tried, I just added some cautionary comments, but if in your tests things are fine, then all good.

pinin4fjords:
Could you clarify why it's necessary to use atlas-bash-util in this way rather than via the conda package? That makes all the contained functions easily accessible (like https://github.com/ebi-gene-expression-group/atlas-bash-util#generic-routines), without all the 'generic routines' imports you have here.

pcm32:
On one side I don't mind the submodule so much since it makes development easier and this is a rather simple package. Otherwise we need to make a conda release each time, rebuild environments, etc.

pcm32:
It is true though that this package should be fairly static.

Snakefile
pmb59: The variables are provisioned as before. Specifically (see below) for 
```
geneNameFile=$( get_geneNameFile_given_organism {params.organism}  )
```
`find_properties_file` now is sourced from atlas-bash-util/generic_routines.sh

Snakefile-sorting-hat
pmb59: I have checked that all the code being replaced is equal. It works

bin/quantile_normalize.sh
pmb59: In this case, they are identical.

Snakefile
pmb59: I did, however, found another case where a variable needed to be exported [c2e555f](https://github.com/ebi-gene-expression-group/bulk-recalculations/pull/40/commits/c2e555f44c70b977d64ab14ee591da123e55be6f)

pmb59:
> Could you clarify why it's necessary to use atlas-bash-util in this way rather than via the conda package? That makes all the contained functions easily accessible (like https://github.com/ebi-gene-expression-group/atlas-bash-util#generic-routines), without all the 'generic routines' imports you have here.

@pinin4fjords In this case, currently, a submodule is more of a benefit (as @pcm32 mentioned, it makes development easier) for this collection of bash functions, instead of using dependency management system. The submodule implies some imports but it is also true that arranging a self-contained conda env in several Snakamake rules also brings a certain computational cost associated to it.  Nevertheless, I will open an issue to consider moving to conda at later stage.

pcm32 (approved):
LGTM - I would add peach API as config, but since we are changing that very soon I'm not so concerned.

Issue #38, opened 2022-05-20
Refactor scripts using a submodule of the atlas-analysis repo


pmb59:
Consider adding [atlas-analysis](https://github.com/ebi-gene-expression-group/atlas-analysis) as a submodule

PR #43, opened 2022-06-06
Refactor analysis scripts using atlas-analysis submodule

pmb59:
The refactoring was motivated the issue https://github.com/ebi-gene-expression-group/bulk-recalculations/issues/38, and it has been done in concert with this [PR](https://github.com/ebi-gene-expression-group/atlas-analysis/pull/6).


pcm32:
I don't see anything about the submodule for atlas-analysis here though, maybe you are missing a commit where you added that?

pmb59:
> I don't see anything about the submodule for atlas-analysis here though, maybe you are missing a commit where you added that?

@pcm32 This was WIP. It is now ready for ready for review, after passing all my tests in codon.

Snakefile
```diff
@@ -16,7 +16,6 @@ def read_metadata_summary():
 
 read_metadata_summary()
 
-
 def get_methods_template_baseline():
```
irisdianauy, line 19: Perhaps this is not part of the intent of the PR, but I think this function (and other similar functions below) can be replaced by the directory `get` function (since `config` is a directory object). So something like `config.get("param_name", None)` inline where the parameter values are needed, in which case there would be no need to define several functions.  E.g., in the case of `get_from_config_or_metadata_summary(label)`, it could be `config.get(label, metadata_summary[label])`.

irisdianauy (approved):
Aside from the minor comment on functions, LGTM if Pablo's comment has been addressed.

Snakefile
pmb59: thank you @irisdianauy for the suggestion. I've opened a new issue https://github.com/ebi-gene-expression-group/bulk-recalculations/issues/48 and it will be handled in a different PR

pmb59:
@irisdianauy Thanks for reviewing!

PR #46, opened 2022-06-25
set minimum snakemake version

pmb59:
Add minimum required version of Snakemake.


pcm32:
LGTM, I don't get an approval button on the phone.

PR #49, opened 2022-08-12
remove arg in rnaseqQC.sh

pmb59:
After the last commit here https://github.com/ebi-gene-expression-group/atlas-analysis/pull/6 we do not need to provide this parameter upstream


irisdianauy (approved):
LGTM!

PR #61, opened 2023-01-10
Get rid of Peach API use in Atlas bulk analysis

pmb59:
As the peach API is down, retrieval of experiment privacy type (public or private) is now done with an internal file provided by the bio studies team.

In addition, enable RAM management in other rules, and increase memory in last attempt to max allowed (300 GB).

### DIFF
--- a/bin/reprocessing_routines.sh
+++ b/bin/reprocessing_routines.sh
@@ -123,19 +123,7 @@ get_biostudies_privacy_status() {
 
     if [ $exp_import == "MTAB" ]; then
 
-        statusStudy=""
-
-        # Use file provided by Biostudies to get privacy status, if it exists.
-        if [[ -f "$privStatusFile" ]]; then
-            statusStudy=$(awk -v p="$expAcc" '$1 == p {print $2}' $privStatusFile)
-        fi
-
-        # If the file doesn't exist, or exists but the above conditional
-        # returns an empty string, try the Biostudies api
-        if [[ -z "$statusStudy" ]]; then
-            statusStudy=$( get_biostudies_api_info $expAcc "isPublic" )
-            if [[ -z "$statusStudy" ]]; then statusStudy="false"; fi
-        fi
+        statusStudy=$(awk -v p="$expAcc" '$1 == p {print $2}' $privStatusFile)
 
         if [ "$statusStudy" == "true" ]; then
             privacyStatus="public"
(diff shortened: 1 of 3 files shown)


bin/reprocessing_routines.sh
```diff
@@ -112,9 +112,38 @@ get_magetab_for_experiment() {
 
 # 7: see directory, see its contents, can write (fg_atlas only)
 # 5: see directory, see its contents: public read-only directory
+
+## get privacy status for any experiments
+## -MTAB- experiments loaded by biostudies-AE/Annotare uis checked internal file
+## -GEOD-/-ERAD-/-ENAD- are loaded as public from now on
+get_biostudies_privacy_status() {
```
irisdianauy, line 119: Not sure which rule this would be used in, but I saw that in some rules, the `generic_routines.sh` script is sourced along with `reprocessing_routines.sh` (as [here](https://github.com/ebi-gene-expression-group/bulk-recalculations/blob/e4f47498c9d2aaac9f28008c2f749b394d136ede/Snakefile#L1117-L1118)) in which case redefining this function will no longer be needed when [this PR](https://github.com/ebi-gene-expression-group/atlas-prod/pull/254) is merged.

bin/reprocessing_routines.sh
```diff
@@ -112,9 +112,38 @@ get_magetab_for_experiment() {
...
+        fi
+    else
+        # if not MTAB, ie. GEOD or ENAD or ERAD are all loaded as public
+        privacyStatus="public"
+    fi
+
+    echo $privacyStatus
+}
+
 experiment_directory_permissions_from_peach_api_privacy() {
```
irisdianauy: Perhaps a name change to remove the reference to the peach API? In this repo, the function is only used here in this script.  The same function is renamed `experiment_directory_permissions_from_biostudies_privacy` in the atlas-prod repo in [this PR](https://github.com/ebi-gene-expression-group/atlas-prod/pull/254).

bin/reprocessing_routines.sh
pmb59: The example you mentioned
```
source {workflow.basedir}/atlas-bash-util/generic_routines.sh
```
comes from local submodule `atlas-bash-util` while the PR you mention if from `atlas-prod`

pmb59:
@irisdianauy  I have rolled back to the case where we use only the bio studies file for data production processes. This has been recommended by the BioStudies team, which updates the file every 30 min

bin/reprocessing_routines.sh
irisdianauy: Oh sorry, I mis-traced.  LGTM.

irisdianauy:
LGTM

irisdianauy (approved):

PR #92, opened 2026-07-07
update DE  plots

anilthanki:


irisdianauy (approved):
Please pin new libs to version, as copilot suggested. Other than that, looks okay to me.

PR #93, opened 2026-07-13
Fix vulnerabilities

anilthanki:


Snakefile
```diff
@@ -9,6 +10,28 @@ os.makedirs('logs', exist_ok='True')
 
 metadata_summary = {}
 
+SAFE_IDENTIFIER_PATTERN = re.compile(r"^[A-Za-z0-9_.-]+$")
```
irisdianauy, line 13: What type of string identifiers do you intend to match with this pattern?

anilthanki: Letters, numbers and `_`, `.`, `-` no other special characters

Snakefile
```diff
@@ -370,7 +413,12 @@ ruleorder: decorate_differential_rnaseq > decorate_differential_proteomics
 
 wildcard_constraints:
     accession="E-\D+-\d+",
-    metric="tpms|fpkms"
+    metric="tpms|fpkms",
+    contrast_id="[A-Za-z0-9_.-]+",
+    assay_id="[A-Za-z0-9_.-]+",
+    array_design="[A-Za-z0-9_.-]+",
+    gid="[A-Za-z0-9_.-]+",
+    assay="[A-Za-z0-9_.-]+"
```
irisdianauy, line 421: Would these generic patterns suffice or would more id-specific patterns be applicable? If no, please make them more specific.

Snakefile
```diff
@@ -370,7 +413,12 @@ ruleorder: decorate_differential_rnaseq > decorate_differential_proteomics
 
 wildcard_constraints:
     accession="E-\D+-\d+",
```
irisdianauy: To be more specific with accessions:
```suggestion
    accession=r"E-[A-Z]+-\d+",
```

Snakefile
```diff
@@ -321,7 +363,8 @@ def get_array_design_from_xml(wildcards):
     itemlist = xmldoc.getElementsByTagName('array_design')
     array_designs_grabbed = []
     for s in itemlist:
-        array_designs_grabbed.append( " ".join(s.firstChild.nodeValue.split() ) )
+        array_design = " ".join(s.firstChild.nodeValue.split())
+        array_designs_grabbed.append(validate_identifier(array_design, "array design"))
```
irisdianauy, line 367: What happens when `validate_identifier` raises an error? Does the script end, or does an invalid value get appended to `array_designs_grabbed`?

anilthanki: script ends

